PR #2, opened 2015-08-31
CHMS-189 | Adding defaults for 'filters' parameter

abarriosr:


abarriosr:
New method purge() has been added to the library.

src/ContentHub.php
```diff
@@ -214,6 +214,23 @@ public function deleteEntity($uuid)
     }
 
     /**
+     * Purges all entities from the Content Hub.
+     *
+     * This method should be used carefully as it deletes all the entities for
+     * the current subscription from the Content Hub.
+     */
+    public function purge() {
+        $list = $this->listEntities();
```
scor: just a question. looks like we have 4 spaces here for indentation. Are we using a different coding standard for this PHP client library?

abarriosr: Yeap, this is not Drupal. We are using http://www.php-fig.org/psr/psr-2/

scor: ok that is fine. what tool do you use to follow this coding standard? do you have an IDE and a code style checker?

abarriosr: I used to have a special settings configured in PHP Storm but that blew up when my laptop died. I will have to set this up again.

PR #3, opened 2015-08-31
Updated to guzzle 6 and the updated hmac library

lkhaas:
Would have to remove the repositories change to the composer.json file once that update has been made.... but in the meantime - here's one way it could work to use Guzzle 6 here.  Bring on the comments!


ynx:
This pull request is rejuvenated by pull request 7. Closing.

PR #6, opened 2016-02-19
CHMS-529 | Empty fields are not sent to the hub which causes issue in re-syncing

charuag:


sumeetpareek:
Tests need to be updated too, now that we are allowing NULL values.

PR #11, opened 2016-04-20
CHMS-110 | PHP Lib passes extra client and lib info in user-agent header

sumeetpareek:


src/ContentHub.php
```diff
@@ -12,6 +12,10 @@
 
 class ContentHub extends Client
 {
+    // Override VERSION inherited from GuzzleHttp::ClientInterface
+    const VERSION = '0.6.4';
```
scor: Just a note to myself: we will need to update this VERSION number every time we make a new release of this PHP library... and that will also include when we release this code 🎯

cpliakas: Any way to automate this? https://github.com/sebastianbergmann/version seems like it might help, it has 14M installs so other projects are finding it useful.

abarriosr: This package looks cool. It adds the git commit into the version using the most recent tag. Wouldn't be hard to add it I think. It could take over the decision to set the version information for us. /me likes it...

cpliakas: Great. I'll post a separate story for that piece, shouldn't hold up this PR.

sumeetpareek:
`master` branch now uses guzzle v6.0+ which is needed for D8 .. this is NOT what we ship to the D7 Content Hub clients.

`0.6.x` is the branch were we maintain what is shipped to the D7 customers. 

So all this code should have been added on top of the later, which is done here - https://github.com/acquia/content-hub-php/pull/14 

Closing this one.

PR #14, opened 2016-04-22
CHMS-110 | On top of 0.6.x branch, PHP Lib passes its own info in user-agent header, and accepts info from clients modules too

sumeetpareek:


sumeetpareek:
Sticking with the convention of keeping the branch name limited to `CHMS-xxx` and closing this PR in favor of - https://github.com/acquia/content-hub-php/pull/15

scor:
Thanks @sumeetpareek

PR #22, opened 2017-03-24
LAT-6 Publish from D8 and import from D7.

ynx:


src/Data/Formatter/Localizer/Drupal7.php
```diff
@@ -46,7 +46,9 @@ protected function localizeEntity($data)
                         // Localize Link fields - Hack.
                         // @TODO: Fix this for Links Type.
                         if (isset($fieldValue['uri'])) {
-                            $fieldValue['url'] = $fieldValue['uri'];
+                            // Checking if it is an internal URL. If it is,
+                            // then strip out the 'internal:/' prefix.
+                            $fieldValue['url'] = (substr($fieldValue['uri'], 0, 10) === 'internal:/') ? str_replace('internal:/', '', $fieldValue['uri']) : $fieldValue['uri'];
```
varvashenya: @abarriosr I tested an internal link functionality locally. It works but it creates links at destination site using nids from source site. I think we should convert internal links into external replacing 'internal:/' part with source site uri.
```
'internal:/node/10' => 'http://source.site/node/10'
```

src/Attribute.php
```diff
@@ -194,4 +194,58 @@ public function getTypeHandlers()
...
+                break;
+
+            case self::TYPE_ARRAY_REFERENCE:
+                $this['type'] = self::TYPE_REFERENCE;
+                break;
+
+            default:
+                // If this attribute is not a type array, then return as it is.
+                return $this;
+        }
```
ynx: Can we bank on the fact that the multiple is always going to be "array<XXXXXXX>", and just regex-strip the surrounding "array<...>"? Tested this could work:

```$attributeType = 'array<boolean>';
$singularAttributeType = preg_replace('/^array\<|\>$/', '', $attributeType);

if ($attributeType === $singularAttributeType) {
    return $this;
}

echo $singularAttributeType;
```

(warning: not tested)

I have no problem this code as-is, though.

PR #24, opened 2017-04-20
LAT-362 Refactor, fix, and tests.

ynx:


src/Data/Formatter/Transformer/Entity/Drupal7.php
```diff
@@ -5,11 +5,20 @@
...
+    /**
+     * Transform - array string to string.
+     *
+     * @param mixed $data Data to transform
+     * @param mixed $index Transform data of index
+     */
     public function arrayStringToString(&$data, $index)
     {
-        if (!isset($data[$index])) {
+        if (!isset($data[$index]) || empty($data[$index]['type']) || empty($data[$index]['value'])) {
```
ynx: @abarriosr, is this line update correct? Basically, if the entry doesn't already have 'type' and 'value', don't update it?

PR #27, opened 2017-05-02
LAT-35 | Enable Content Hub developers to diagnose and fix issues without engineering intervention

abarriosr:


src/ContentHub.php
```diff
@@ -12,15 +12,24 @@
...
      *
      * @param string $apiKey
      * @param string $secretKey
      * @param string $origin
      * @param array  $config
+     * @param string $api_version
      */
-    public function __construct($apiKey, $secretKey, $origin, array $config = [])
+    public function __construct($apiKey, $secretKey, $origin, array $config = [], $api_version = NULL)
     {
```
matthewmorley: style question: any reason to have $api_version be NULL... instead of type declaring as a string and defaulting it to'v1'. Then you could blindly assign on line 45 without the !empty check.

matthewmorley:
Other than a small style question,  changes look good. Haven't tried/built with it yet.

src/ContentHub.php
abarriosr: Changed. Thanks.

abarriosr: Did the same in the D7 version.

src/ContentHub.php
```diff
@@ -216,25 +234,95 @@ public function updateEntities($resource)
...
+        $endpoint = "/{$this->api_version}/entities/purge";
+        $request = new Request('POST', $endpoint, [], '');
+        return $this->getResponseJson($request);
+    }
+
+    /**
+     * Restores the state of entities before the previous purge.
+     *
+     * Only to be used if a purge has been called previously. This means new
+     * entities added after the purge was enacted will be overwritten by the
```
scor: 'enact' sounds so formal and official. where is that language from?

PR #33, opened 2017-10-09
LAT-1097 | order log results by timestamp by default

scor:


chilic:
@scor This PR looks okay for me.
How you think, should we update Drupal module
https://github.com/acquia/content-hub-d8/blob/8.x-1.x/acquia_contenthub.drush.inc#L768

scor:
I didn't look at the drush command, but only modified the PHP client (this PR) and it worked for me. I don't think we need to change the drush command.

matthewmorley:
Code change looks good.

PR #34, opened 2017-10-11
LAT-1003: Upgrade symfony components for 8.4, ensure 8.3 compatibility.

Infantys:


Infantys:
@scor: I thought about your question and completely agree with lcatlett. For tests, create two branches, one for versions of Symfony 2, the other for version 3. 

Note that Symfony 3 works with phpunit: "> = 4.8.35 <5" (as in Drupal 8.4). And Symfony 2 is enough to run phpunit: "~ 3.0" (as in Drupal 8.3). But I would recommend always using phpunit: "> = 4.8.35 <5" (as in Drupal 8.4).

PR #42, opened 2018-04-05
LCH-1570: Update Hmac Libraries to take Hmacv2 functions.

japerry:


japerry:
HMACv2 is only going to be in Version 2 of the library.

PR #46, opened 2018-07-27
2.x

japerry:


src/CDFAttribute.php
```diff
@@ -52,17 +52,15 @@ class CDFAttribute {
    *
    * @throws \Exception Unsupported data type exception.
    */
-  public function __construct($id, $type, $value = NULL, $language = 'und') {
+  public function __construct($id, $type, $value, $language = 'und') {
```
EclipseGc: This whole file needs to stay the same, so it shouldn't be included in this diff.

src/CDFObject.php
```diff
@@ -120,7 +120,7 @@ public function getAttribute($id) {
   }
 
   public function addAttribute($id, $type, $value, $language = 'und', $class = '\Acquia\ContentHubClient\CDFAttribute') {
-    if ($class != '\Acquia\ContentHubClient\CDFAttribute' && !is_subclass_of($class, '\Acquia\ContentHubClient\CDFAttribute')) {
+    if (!is_subclass_of($class, '\Acquia\ContentHubClient\CDFAttribute')) {
```
EclipseGc: This file also should not be touched by the diff.

EclipseGc:
Manually merged from a clean copy.

PR #58, opened 2019-02-21
LCH-2263 | Fixing existing and add new tests.

dougan88:
@EclipseGc @japerry 
https://github.com/acquia/content-hub-php/blob/a68f22d806e61e253d60304ab601f3422ecff8dc/src/CDFDocument.php#L14

Is it necessary to use ...$entities (triple dot) operator in the line above? It causes hhvm travis ci tests failure.


src/CDFDocument.php
```diff
@@ -20,13 +20,13 @@ public function getEntities() {
   }
 
   public function getCDFEntity($uuid) {
-    if ($this->hasEntity($uuid)) {
-      return $this->entities[$uuid];
-    }
+    return ($this->hasEntity($uuid)) ? $this->entities[$uuid] : null;
```
c0d39uru: 1- Parentheses around the condition is redundant
2- This code could be simplified as:
``` php
return $this->entities[$uuid] ?? null;
```

src/CDFDocument.php
```diff
@@ -20,13 +20,13 @@ public function getEntities() {
...
   }
 
   public function setCDFEntities(CDFObject ...$entities) {
-    unset($this->entities);
+    //unset($this->entities);
+    //With unset() if $entities list is empty, hasEntities() will give an error
+    $this->entities = [];
     foreach ($entities as $entity) {
       $this->entities[$entity->getUuid()] = $entity;
     }
```
c0d39uru, line 30: Setting class member in a loop is very inefficient. In order to fix all issues with this code simply use a local array and in the end set the class member to this array.

test/CDFDocumentTest.php
```diff
@@ -0,0 +1,248 @@
...
+class CDFDocumentTest extends PHPUnit_Framework_TestCase
+{
+    /**
+     * @var CDFDocument
+     */
+    protected $cdfDocument;
+
+    public function setUp()
+    {
+        parent::setUp();
```
c0d39uru: I would get rid of calling parent setup. I don't know of a use case for it.

test/CDFDocumentTest.php
```diff
@@ -0,0 +1,248 @@
...
+        $this->cdfDocument = new CDFDocument();
+    }
+
+    public function tearDown()
+    {
+        parent::tearDown();
+        unset($this->cdfDocument);
```
c0d39uru: Same for parent::tearDown

test/CDFDocumentTest.php
```diff
@@ -0,0 +1,248 @@
...
+     * @param $objectToAdd
+     * @param $notAddedObject
+     */
+    public function testHasEntities($objectToAdd, $notAddedObject)
+    {
+        //Check hasEntities method before and after we add an Object
+        $this->assertEquals($this->cdfDocument->hasEntities(), false);
+        $this->cdfDocument->addCDFEntity($objectToAdd);
+        $this->assertEquals($this->cdfDocument->hasEntities(), true);
+    }
```
c0d39uru: The AAA (Arrange, Act, Assert) pattern is a common way of writing unit tests for a method under test. This test could be split into two.

c0d39uru: Try to use better names for your tests instead of using comments along with your asserts. For instance:
``` php
public function test_hasEntities_returns_false_before_any_object_being_added($objectToAdd, $notAddedObject)
{
        $this->assertEquals($this->cdfDocument->hasEntities(), false);
}
```

test/CDFDocumentTest.php
```diff
@@ -0,0 +1,248 @@
...
+                json_encode(['entities' => []]),
+                json_encode(['entities' => [
+                    $cdfObjectMockFirstToArray,
+                    $cdfObjectMockSecondToArray
+                ]]),
+            ]
+        ];
+    }
+
+}
```
c0d39uru, line 427: There are lots of repetition especially in `Data Providers`. I strongly believe providers are being overused here. I see bunch of loops and logic in tests, which should only contain simple statements. 
Tests should be very easy to understand and follow without much cognitive load.

c0d39uru: In most cases, `willReturn` is a better choice.

test/SettingsTest.php
```diff
@@ -0,0 +1,135 @@
...
+     */
+    public function testSharedSecret($settingsData)
+    {
+        $this->assertEquals($this->settings->getSharedSecret(), $settingsData['sharedSecret']);
+    }
+
+    public function getSettingsData()
+    {
+        return [
+            'data' => [
```
c0d39uru: What is the point in having this key?

test/SettingsTest.php
```diff
@@ -0,0 +1,135 @@
...
+        ];
+    }
+
+    public function settingsDataProvider()
+    {
+        return [
+            $this->getSettingsData()
+        ];
+    }
+
```
c0d39uru, line 134: Why not using getSettingsData as the provider? What is the point in having another method?

src/CDFDocument.php
dougan88: 1. Agree regarding parentheses.
2. I don't know if we should save 5.6 compatibility. If no, you are correct, otherwise we should leave it as is.

src/CDFDocument.php
dougan88: Waiting for the end of your discussion regarding that.

dougan88: If you consider it should be done other way, I will.

test/CDFDocumentTest.php
ynx: Personally I think we should keep `setUp()` and `tearDown()`. `setUp()` can be used to provide universal fixture or trait helpers or etc. Uniformity helps debugging and maintaining those tests easy. I do realize that this is probably not the case here yet, though.

Also I am just providing perspectives. You guys decide which way. ^^

test/CDFDocumentTest.php
ynx: Like `setUp()`, same comments on `tearDown()`.

test/SettingsTest.php
dougan88: The point is to access this data in setUp function and data providers without duplicating it.

src/CDFDocument.php
c0d39uru: I believe for 2.x we assume PHP 7 and up

src/CDFDocument.php
c0d39uru: This has nothing to do with our discussion about not using control structure in tests. Here's my suggestion:

``` php
$result = [];

foreach ($entities as $entity) {
      $result[$entity->getUuid()] = $entity;
}

$this->entities = $result;
```

test/CDFDocumentTest.php
c0d39uru: Like I said, I don't see any use case here, but if it helps uniformity, I am ok with it! Not a big deal!

test/SettingsTest.php
c0d39uru: Doesn't the following work?

``` php
$settingsParameters = $this->getSettingsData();
```

test/SettingsTest.php
c0d39uru: My brain has an obsession towards simplicity, so I am just telling if there is a case for having simpler code, let's take the chance and do it!

test/SettingsTest.php
dougan88: It works, but in my opinion when you read this:
```
$settingsParameters = $this->getSettingsData()['data'];
```
it gives you more understanding of what you get then reading this:
```
$settingsParameters = $this->getSettingsData()[0];
```

c0d39uru: I wonder why you even need `[0]`?

src/CDFDocument.php
dougan88: Oh, sorry, confused this with a different piece of code. Agree here.

test/SettingsTest.php
dougan88: Because then it will be needed to use this dataProvider like this in setUp():
```
$settingsParameters = $this->getSettingsData()[0];
```
Which in my opinion is less ugly to use an additional function.

dougan88: Or duplicate the data from getSettingsData()

test/CDFDocumentTest.php
dougan88: I made it like that to show the different behaviour of the same object.

test/CDFDocumentTest.php
dougan88: Not sure what you mean by lots of repetitions. Where exactly?
willReturn is a better choice than what?

c0d39uru: - As for repetitions, for example there are a few instances of `2014-12-21T20:12:11+00:00Z` or `11111111-0000-0000-0000-000000000000` which, in my opinion, could be centralized.

- Instead of using:
``` php
will($this->returnValue('66666666-0000-0000-0000-000000000000'));
```
You could use:
``` php
willReturn('66666666-0000-0000-0000-000000000000');
```

dougan88: - Yes, it could be moved into a single basic array which will be customized afterwards. Which I think is a good idea. But after all these discussions regarding complicated tests I was not sure whether I should do it like that :) But I will.
- As far as I understand it's the same. But ok, willReturn uses less letters, so I will switch to it.

c0d39uru: No, they are definitely not the same. `will` has more complex stubbing behavior than `willReturn`, which makes it slower.

test/SettingsTest.php
c0d39uru: if you return an array of data in your provider, you definitely need to use index `[0]`, in which case, I . agree that `['data']` is way better. But if your provider simply returns the required data, I don't think you need to use either of them.

dougan88: Provider should return an array of arrays.

test/CDFDocumentTest.php
dougan88: I meant in a context which I'm using it. But what you suggest is a bit better, I agree.

test/SettingsTest.php
c0d39uru: Sorry, I meant `getSettingsData` not the provider.

dougan88: I commited that already in response to your previous comment.

src/CDF/CDFObject.php
```diff
@@ -144,17 +144,26 @@ public function getAttribute($id) {
...
     }
-    $attribute = new $class($id, $type, $value, $language);
+    $attribute = new $className($id, $type, $value, $language);
     $this->attributes[$attribute->getId()] = $attribute;
     // Keep track of the class used for this attribute.
-    if ($class != '\Acquia\ContentHubClient\CDFAttribute') {
-      $this->metadata['attributes'][$attribute->getId()]['class'] = $class;
-    }
-    else {
+    if ($className != CDFAttribute::class) {
```
jabxjab: @dougan88 Please use strict `!==`

src/CDF/CDFObject.php
```diff
@@ -144,17 +144,26 @@ public function getAttribute($id) {
     }
   }
 
-  public function addAttribute($id, $type, $value = NULL, $language = self::LANGUAGE_UNDETERMINED, $class = '\Acquia\ContentHubClient\CDFAttribute') {
-    if ($class != '\Acquia\ContentHubClient\CDFAttribute' && !is_subclass_of($class, '\Acquia\ContentHubClient\CDFAttribute')) {
-      throw new \Exception(sprintf("The %s class must be a subclass of \Acquia\ContentHubClient\CDFAttribute", $class));
+    /**
+     * Factory could be used here in order to stick with Single responsibility principle.
+     *
+     * @param $id
```
jabxjab: @dougan88 Please update phpdoc to include param types and descriptions

src/CDF/CDFObject.php
```diff
@@ -144,17 +144,26 @@ public function getAttribute($id) {
...
+     * @param $type
+     * @param null $value
+     * @param string $language
+     * @param string $className
+     * @throws \Exception
+     */
+  public function addAttribute($id, $type, $value = NULL, $language = self::LANGUAGE_UNDETERMINED, $className = CDFAttribute::class) {
+    if ($className != CDFAttribute::class && !is_subclass_of($className, CDFAttribute::class)) {
```
jabxjab: @dougan88 Please use strict `!==`

test/CDFObjectTest.php
```diff
@@ -0,0 +1,253 @@
...
+            $objectParameters['type'],
+            $objectParameters['uuid'],
+            $objectParameters['created'],
+            $objectParameters['modified'],
+            $objectParameters['origin'],
+            $objectParameters['metadata']
+        );
+    }
+
+    public function tearDown()
```
jabxjab: @dougan88 Would be great to have phpdoc for the method

test/assets/CDFAttributeChild.php
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Acquia\ContentHubClient\assets;
+
+
+use Acquia\ContentHubClient\CDFAttribute;
+
+class CDFAttributeChild extends CDFAttribute
+{
+
+}
```
jabxjab: @dougan88 Missing new line

test/CDFObjectTest.php
```diff
@@ -0,0 +1,253 @@
...
+
+    public function objectDataProvider()
+    {
+        return [
+            [
+                $this->getObjectData()
+            ],
+        ];
+    }
+}
```
jabxjab: @dougan88 Missing new line

test/CDFObjectTest.php
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    public function attributeDataProvider()
+    {
+        return [
+            [
+                'value' => [
+                    'en' => [
+                        6.66,
+                        3.23
```
jabxjab: @dougan88 Missing comma

test/CDFObjectTest.php
```diff
@@ -0,0 +1,253 @@
...
+                    ],
+                    'hu' => [
+                        4.66,
+                        4.23
```
jabxjab: @dougan88 Missing comma

test/CDFObjectTest.php
```diff
@@ -0,0 +1,253 @@
...
+                    ],
+                    CDFObject::LANGUAGE_UNDETERMINED => [
+                        1.22,
+                        1.11
```
jabxjab: @dougan88 Missing comma char

test/CDFObjectTest.php
```diff
@@ -0,0 +1,253 @@
...
+                ],
+            ]
+        ];
+    }
+
+    public function objectDataProvider()
+    {
+        return [
+            [
+                $this->getObjectData()
```
jabxjab: @dougan88 Missing comma

jabxjab (approved):
Please fix minor issues

c0d39uru (approved):

src/CDFDocument.php
```diff
@@ -20,16 +20,17 @@ public function getEntities() {
...
   public function getCDFEntity($uuid) {
-    if ($this->hasEntity($uuid)) {
-      return $this->entities[$uuid];
-    }
+    return $this->entities[$uuid] ?? null;
   }
 
   public function setCDFEntities(CDFObject ...$entities) {
-    unset($this->entities);
+    //unset($this->entities);
```
ynx: Is there a pending commit push? This code shouldn't be here. ^^

test/CDFDocumentTest.php
```diff
@@ -0,0 +1,428 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: eugene
+ * Date: 2/18/19
+ * Time: 3:04 PM
+ */
```
ynx: We don't have a standard to add the line 2 - 7. Before we start this new tradition, we have to get team consensus plus everyone to do the same. For now, please remove line 2 to 7 and stop your IDE from generating this automatically for Content Hub files. ^^

test/CDFDocumentTest.php
```diff
@@ -0,0 +1,428 @@
...
+
+namespace Acquia\ContentHubClient\test;
+
+
```
ynx: Please remove one redundant empty line.

test/CDFDocumentTest.php
ynx: Great. Yes let's keep uniformity.

src/CDFDocument.php
ynx: And line 28 shouldnt be here either?

src/CDFDocument.php
```diff
@@ -20,16 +20,17 @@ public function getEntities() {
...
+    //With unset() if $entities list is empty, hasEntities() will give an error
+    $entitiesList = [];
     foreach ($entities as $entity) {
-      $this->entities[$entity->getUuid()] = $entity;
+        $entitiesList[$entity->getUuid()] = $entity;
```
ynx: Indentation is wrong, should be 2, not 4. (Drupal is really weird to go with 2 ...)

src/CDFDocument.php
ynx: This conversation is resolvable.

ynx:
Left some comments. They don't look too hard to fix.

Great job fixing and adding tests! I really personally enjoyed reading you code. Kudos here sire.

I see the Travis run is not fixed yet, which is in scope. See "1 errored check" on this page. Please fix it before we can merge this code.

ynx:
Separately, "This branch has conflicts that must be resolved". Please solve merge conflict. Please feel free to force-push if required.

src/CDF/CDFObject.php
```diff
@@ -142,17 +144,26 @@ public function getAttribute($id) {
...
+    /**
+     * Factory could be used here in order to stick with Single responsibility principle.
+     *
+     * @param string $id
+     * @param string $type Type of attribute
+     * @param null $value Attribute value
+     * @param string $language
+     * @param string $className Class name of the created object
+     * @throws \Exception
+     */
```
ynx: Wrong indentation.

src/CDF/ClientCDFObject.php
```diff
@@ -12,12 +12,13 @@ class ClientCDFObject extends CDFObject {
...
-   * @param string $uuid
-   * @param array $settings
-   */
+    /**
+     * ClientCDFObject constructor.
+     *
+     * @param string $uuid
+     * @param array $settings
+     * @throws \Exception
+     */
```
ynx: Wrong indentation.

test/CDFDocumentTest.php
ynx: Still pending.

test/CDFDocumentTest.php
ynx: Still pending.

ynx:
Just noticed all the new files are with 4-space indentation, which is inconsistent with other files in this codebase (i.e. 2-space indentation).

Please either fix within this PR by reduce to 2-space indentation, or highlight this required fix in the JIRA ticket that you are about to create to encapsulate next works.

test/CDFDocumentTest.php
dougan88: I guess phpstorm adds two blank lines after namespace. But will do.

dougan88:
@ynx Travis ci is passing for 7.1, but failing for 5.6, which as we discussed should be removed. But it's not clear from our communication whether it should be removed.

ynx:
The only PHP version we are mandatorily testing on is 7.1. Please add 7.2 there as well.

We should not be testing on 7.0 or lower.

ynx (approved):

PR #59, opened 2019-03-06
LCH-2682 | Fix stripping of 'v2' from base_uri by guzzle HTTP lib.

nkarpenko:
Guzzle was stripping the v2 tags from 'base_uri' variable in the config so we were always querying v1 endpoints. Guzzle HTTP lib strips it out by default if no trailing slash is added.

Refer to same issue here: https://stackoverflow.com/questions/30834147/base-uri-not-being-based-from-guzzle-client-instantiation

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -623,6 +623,25 @@ class ContentHubClient extends Client
     }
 
     /**
+     * Add entities to Intrest List.
+     *
+     * @param string $webhook_uuid
+     *   The UUID of the webhook
+     * @param array $uuids
+     *   Entity UUIDs to add to Interest List
+     *
+     * @return \Psr\Http\Message\ResponseInterface
+     *   The response.
+     *
+     * @throws \GuzzleHttp\Exception\RequestException
+     */
+    public function addEntitiesToInterestList($webhook_uuid, $uuids) 
+    {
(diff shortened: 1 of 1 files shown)


jabxjab (approved):

.gitignore
```diff
@@ -4,3 +4,4 @@ vendor
 composer.lock
 
 nbproject
+.idea
```
ynx: @looking-4-justice Can you please implement this `.idea` change to all of our main codebases? I will review them.

list:
* content-hub-d8 (1.x and 2.x)
* content-hub-d7
* content-hub-php (1.x and 2.x)
* content-hub-ember (1.x and 2.x)
* hive

Please and thank you! ^^

ynx: @nkarpenko Can you please implement this `.idea` change to all of our main plexus codebases? I will review them.

list:
* plexus
* plexus-client-go
* customer-auth
* (any other codebases)

Please and thank you! ^^

c0d39uru: By all means. Just a couple of questions:
- Do I need a ticket for this (hence a branch with the same number)?
- Which branch should I branch off of for each project?

ynx: Documenting as discussed:
* No because it is anticipated to be less than 30 mins work.
* Master or develop or equivalent.

PR #64, opened 2019-03-14
LCH-2260: Missing or incomplete DockBlocks [fixed]. Minor CS fixes

Max-Yudin:


jabxjab (approved):

src/ContentHubClient.php
```diff
@@ -340,514 +446,583 @@ public function putEntity(CDFObject $object)
...
+   *
+   * @throws \Exception
+   */
+  public function getWebHook($url) {
+    $webhooks = $this->getWebHooks();
+    foreach ($webhooks as $webhook) {
+      if ($webhook['url'] == $url) {
+        return $webhook;
+      }
     }
```
c0d39uru: I think we can do this without foreach just by using array functions.

src/ContentHubClient.php
```diff
@@ -340,514 +446,583 @@ public function putEntity(CDFObject $object)
...
+   *   The UUID of the client to delete, blank for current client.
+   *
+   * @return \Psr\Http\Message\ResponseInterface
+   *   Response.
+   *
+   * @throws \GuzzleHttp\Exception\RequestException
+   */
+  public function deleteClient($client_uuid = NULL) {
+    $settings = $this->getSettings();
+    $uuid = isset($client_uuid) ? $client_uuid : $settings->getUuid();
```
``` php
    $uuid = $client_uuid ?? $settings->getUuid();
```

src/ContentHubClient.php
```diff
@@ -340,514 +446,583 @@ public function putEntity(CDFObject $object)
...
+   *
+   * @throws \GuzzleHttp\Exception\RequestException
+   * @throws \Exception
+   */
+  public function getFilterByName($filter_name) {
+    $result = $this::getResponseJson($this->get("filters?name={$filter_name}"));
+    if ($result['success'] == 1) {
+      return $result['data'];
     }
+    return NULL;
```
``` php
return $result['success'] === 1 ? $result['data'] : NULL;
```

src/ContentHubClient.php
Max-Yudin: Agree with you but code refactoring is out of scope of this task.

src/CDF/CDFObject.php
```diff
@@ -52,13 +73,26 @@ class CDFObject implements CDFObjectInterface {
...
+   *   Modified date.
    * @param string $origin
-   * @param array $metadata
-   */
-  public function __construct($type, $uuid, $created, $modified, $origin, $metadata = []) {
+   *   Origin UUID.
+   * @param array|mixed $metadata
+   *   Metadata.
+   */
+  public function __construct(
```
EclipseGc: Let's not do constructors this way. I think it's much more difficult to read and it doesn't match any standard we follow.

src/Asset.php
```diff
@@ -2,67 +2,86 @@
...
-    public function setUrl($url)
-    {
-        $this['url'] = $url;
+  /**
+   * Asset constructor.
+   *
+   * @param array $array
+   *   Asset data.
+   */
+  public function __construct(array $array = []) {
```
EclipseGc: So the { on the same line is a drupal standard, but seeing as how this is a php library we're going to follow the standard the rest of the php community uses which is { on a separate line. Let's change that back.

src/CDF/CDFObject.php
```diff
@@ -145,29 +195,43 @@ public function getAttribute($id) {
...
-    if ($className !== CDFAttribute::class && !is_subclass_of($className, CDFAttribute::class)) {
-      throw new \Exception(sprintf("The %s class must be a subclass of \Acquia\ContentHubClient\CDFAttribute", $className));
+  public function addAttribute(
+    $id,
+    $type,
+    $value = NULL,
+    $language = self::LANGUAGE_UNDETERMINED,
+    $className = CDFAttribute::class
+  ) {
+    if ($className !== CDFAttribute::class && !is_subclass_of(
```
EclipseGc: Same with if statements. Single line is usually much more readable. Let's keep it that way for the time being.

src/ContentHubClient.php
```diff
@@ -18,209 +18,308 @@
...
-     * @since 0.2.0
-     */
-    public function ping()
-    {
-        $config = $this->getConfig();
-        // Create a new client because ping is not behind hmac.
-        $client = new Client(['base_uri' => self::makeBaseURL($config['base_uri'])]);
-        return $this->getResponseJson($client->get('ping'));
+    // Setting up the User Header string.
+    $user_agent_string = self::LIBRARYNAME . '/' . self::VERSION . ' ' . \GuzzleHttp\default_user_agent(
```
EclipseGc: This is super weird looking with the closing parentheses on the next line.

src/ContentHubClient.php
```diff
@@ -18,209 +18,308 @@
...
+    // "base_url" parameter changed to "base_uri" in Guzzle6, so the following
+    // line is there to make sure it does not disrupt previous configuration.
+    if (!isset($config['base_uri']) && isset($config['base_url'])) {
+      $config['base_uri'] = self::makeBaseURL(
+        $config['base_url'],
+        $api_version
+      );
+    }
+    else {
+      $config['base_uri'] = self::makeBaseURL(
```
EclipseGc: Multilining things like this doesn't make them more readable.

src/ContentHubClient.php
```diff
@@ -18,209 +18,308 @@
...
+   *
+   * @throws \GuzzleHttp\Exception\RequestException
+   * @throws \Exception
+   *
+   * @since 0.2.0
+   */
+  public function ping() {
+    $config = $this->getConfig();
+    // Create a new client because ping is not behind hmac.
+    $client = new Client(
```
EclipseGc: same here, let's do this as a single line.

EclipseGc (changes requested):
I didn't review the whole diff because I found enough I wanted to see changed through out. So let's address that stuff and I'll re-review.

japerry:
This set of code changes looks good. committing.

PR #65, opened 2019-03-27
LCH-2727: Interest list operations - error messages [added]

jabxjab:


saxumVermes (changes requested):
What if the Exception is type of RequestException?

src/ContentHubClient.php
```diff
@@ -1032,6 +1032,16 @@ protected function getExceptionMessage($method, array $args, \Exception $excepti
...
+                case 'addEntitiesToInterestList':
+                    return $this->getErrorResponse($response->getStatusCode(),
+                      sprintf('Error trying to add entities to the interest list (Error Code = %d: %s)',
+                        $response->getStatusCode(), $response->getReasonPhrase()));
+
+                case 'deleteInterest':
+                    return $this->getErrorResponse($response->getStatusCode(),
+                      sprintf('Error trying to remove entity from the interest list (Error Code = %d: %s)',
+                        $response->getStatusCode(), $response->getReasonPhrase()));
+
```
c0d39uru: I think it's worth doing a small refactor here (if possible). I can see a repetitive pattern which could be abstracted out to make the code DRYer. AFAICS, each case is returning the exact same thing except for the message. I think we can benefit from the following pattern to reduce clutter:

``` php
switch ($decision) {
   case 'pattern-1':
      $message = 'some message goes here';
      break;
   case 'pattern-2':
      $message = 'some other message';
      break;
    ....
}

return $this->getErrorResponse(
   $response->getStatusCode(),
   $message,
   $response->getStatusCode(), $response->getReasonPhrase())
);
```

jabxjab: @looking-4-justice I would say that this could be done in the scope of https://backlog.acquia.com/browse/LCH-2722, not the current one.

c0d39uru: Agreed!

jabxjab: @looking-4-justice Actually, I've made a refactoring because of @saxumVermes mentioned `RequestException`. The solution was looking so horrible (too many duplications), so I decided to refactor this right now.

src/ContentHubClient.php
```diff
@@ -955,127 +955,106 @@ public function __call($method, $args)
    */
     protected function getExceptionMessage($method, array $args, \Exception $exception)
     {
+        $response = NULL;
+        $status_code = 500;
```
c0d39uru: Could you please use constants defined in Response class instead of 500?

src/ContentHubClient.php
```diff
@@ -955,127 +955,106 @@ public function __call($method, $args)
...
-                default:
-                    return $this->getErrorResponse(500,
-                      sprintf('Error trying to connect to the Content Hub. Are your credentials inserted correctly? (Error Message = %s)',
-                        $exception->getMessage()));
-            }
+
+        switch ($method) {
+            case 'getClientByName':
+                // All good, means the client name is available.
+                if ($response && 404 === $status_code) {
```
c0d39uru: And here instead of 404

src/ContentHubClient.php
```diff
@@ -955,127 +955,106 @@ public function __call($method, $args)
...
+
+            case 'addWebhook':
+                $message = sprintf('There was a problem trying to register Webhook URL = %s. Please try again. (Error Code = %d: %s)',
+                    $args[0], $status_code, $error_details);
+                break;
+
+            case 'deleteWebhook':
+                // This function only requires one argument (webhook_uuid), but
+                // we are using the second one to pass the webhook_url.
+                $webhook_url = isset($args[1]) ? $args[1] : $args[0];
```
``` php
                $webhook_url = $args[1] ?? $args[0];

```

src/ContentHubClient.php
```diff
@@ -955,127 +955,106 @@ public function __call($method, $args)
...
+                $message = sprintf('There was a problem trying to <b>unregister</b> Webhook URL = %s. Please try again. (Error Code = %d: @%s)',
```
c0d39uru: As far as I know, it's a long time that using `b` to make the text bold is deprecated (same goes true for `i` and alike). Not sure if we can avoid using it here, too.

src/ContentHubClient.php
jabxjab: @looking-4-justice Don't see constants in Response class

jabxjab: @looking-4-justice Nevermind... found in Symfony component, not a Guzzle class

jabxjab:
@looking-4-justice @saxumVermes Updated, please review

c0d39uru (approved):

saxumVermes (approved):

Ateee89 (approved):

abarriosr:
This branch was based on an older version of the library.  Closing this PR in favor of this one https://github.com/acquia/content-hub-php/pull/65.

PR #67, opened 2019-03-29
Filter secret out of the clientcdf object.

japerry:


kaynenh:
This was addressed differently after this PR was created. Closing.

PR #69, opened 2019-04-04
LCH-2481: Fix phpdoc and ping method.

saxumVermes:


src/ContentHubClient.php
```diff
@@ -102,7 +102,7 @@ public function ping()
     {
         $config = $this->getConfig();
         // Create a new client because ping is not behind hmac.
-        $client = new Client(['base_uri' => self::makeBaseURL($config['base_uri'])]);
+        $client = new Client(['base_uri' => $config['base_url']]);
         return $this->getResponseJson($client->get('ping'));
```
c0d39uru, line 106: What was wrong with `self::makeBaseURL`? Because makeBaseURL is there to make sure that base URL will always end with slash and by removing it from the equation, you're compromising it.

looking-4-justice: makeBaseURL is there to make sure that base URL will always end with slash and by removing it from the equation, you're compromising it.

saxumVermes: Fixed

c0d39uru (approved):

Ateee89 (approved):

PR #72, opened 2019-04-24
LCH-2661: Add missing tests to 2.x

dougan88:


test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace Acquia\ContentHubClient\test;
+
```
jabxjab, line 4: @dougan88 Excessive blank line

test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
...
+class ClientCDFObjectTest extends TestCase
+{
+  /**
+   * @var ClientCDFObject
+   */
+  private $clientCdfObject;
+
+  /**
+   *
```
jabxjab: @dougan88 Please fill in all PHPDoc block

test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
...
+  public function setUp() : void
+  {
+    parent::setUp();
+    $settingsData = $this->getSettingsData();
+    try {
+      $this->clientCdfObject = ClientCDFObject::create(
+        'client_cdf_id_1',
+        $settingsData
+      );
+    } catch (\Exception $exception) {
```
jabxjab: @dougan88 What is the purpose of such a catch?

test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  /**
+   *
+   */
+  public function tearDown() :void
```
jabxjab: @dougan88 Please follow a consistent code styling `:void` vs `: void` (line 46)

test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
...
+        'sharedSecret' => null,
+        'webhook' => [
+          'http://example1.com/webhooks' => '00000000-0000-0000-0000-000000000000',
+          'http://example2.com/webhooks' => '11111111-0000-0000-0000-000000000000',
+        ],
+      ],
+    ];
+  }
+
+}
```
jabxjab: @dougan88 Missing new line

test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
...
+   */
+  public function testGetWebhook() : void
+  {
+    $this->assertEquals($this->getSettingsData()['settings']['webhook'], $this->clientCdfObject->getWebhook());
+  }
+
+  public function getSettingsData() : array
+  {
+    return [
+
```
jabxjab: @dougan88 excessive blank line

test/GetCDFTypeEventTest.php
```diff
@@ -0,0 +1,98 @@
...
+   * @return array
+   */
+  private function getCdfTypeEventData() : array
+  {
+    return [
+      'type' => 'dummy_type',
+    ];
+  }
+
+}
```
jabxjab: @dougan88 missing new line

test/ClientCDFObjectTest.php
dougan88: @jabxjab the reason why this is made like that is because phpstorm highlights it as an unhandled exception. Although could print an error message here, but in my opinion this is optional since we're only going to run this in dev environment.

test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
...
+    parent::setUp();
+    $settingsData = $this->getSettingsData();
+    try {
+      $this->clientCdfObject = ClientCDFObject::create(
+        'client_cdf_id_1',
+        $settingsData
+      );
+    } catch (\Exception $exception) {
+    }
+  }
```
c0d39uru, line 28: As a rule of thumb, no control structure (if/else/try/catch) is allowed in tests. The static code analyzers either don't know this or they are not context-aware. 
In case you wanna see if your test raises any exception, **expectException** is your friend.

test/ClientCDFObjectTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   *
+   */
+  public function testGetClientName() : void
+  {
+    $this->assertInstanceOf(CDFAttribute::class, $this->clientCdfObject->getClientName());
+    $this->assertEquals(CDFAttribute::TYPE_STRING, $this->clientCdfObject->getClientName()->getType());
+    $this->assertEquals($this->getSettingsData()['settings']['name'], $this->clientCdfObject->getClientName()->getValue()[CDFObject::LANGUAGE_UNDETERMINED]);
```
c0d39uru: Use caching to prevent repetitive calls to this->clientCdfObject->getClientName().

test/ContentHubClientTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   *
+   */
+  public function tearDown() : void
+  {
+    parent::tearDown();
+    unset($this->contentHubClient);
+    \Mockery::close();
+    gc_collect_cycles();
```
c0d39uru: You shouldn't be worried about gc.

test/ContentHubClientTest.php
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public function testDefinition()
+  {
+    $successJsonResponse = \GuzzleHttp\json_encode([
+      'response_code' => 200,
```
c0d39uru: Use constants instead of pure http response codes.

test/ContentHubClientTest.php
```diff
@@ -0,0 +1,82 @@
...
+    ]);
+    $this->contentHubClient->shouldReceive('getResponseJson')
+      ->andReturn($successJsonResponse);
+    $responseMock = \Mockery::mock(Response::class);
+    $this->contentHubClient->shouldReceive('request')
+      ->andReturn($responseMock);
+
+    try {
+      $this->assertEquals($successJsonResponse, $this->contentHubClient->definition('dummyendpoint'));
+      $this->assertNotEquals($failureJsonResponse, $this->contentHubClient->definition('dummyendpoint'));
```
c0d39uru: The second assertion is useless given that successJsonResponse and failureJsonResponse are not equal.

test/ClientCDFObjectTest.php
jabxjab: @dougan88 Please use `@throws \Exception` phpdoc instead

test/ClientCDFObjectTest.php
dougan88: @jabxjab I'm not sure what needs to be placed there since they don't accept or return any parameters. And also they have self-explanatory names. So the only thing I can do is basically duplicate functions' names.

jabxjab: @dougan88 please refer to psr-1/2
In this particular case `{@inheritdoc}` would be enough

test/ClientCDFObjectTest.php
dougan88: fixed

test/ClientCDFObjectTest.php
dougan88: fixed

test/ClientCDFObjectTest.php
dougan88: switched to @throws

dougan88: switched to \@throws

dougan88: switched to `@throws`

test/ClientCDFObjectTest.php
dougan88: fixed

test/ClientCDFObjectTest.php
dougan88: fixed

test/ClientCDFObjectTest.php
dougan88: fixed

test/GetCDFTypeEventTest.php
dougan88: fixed

test/ClientCDFObjectTest.php
dougan88: switched to `@throws` in comments

test/ClientCDFObjectTest.php
dougan88: fixed

test/ContentHubClientTest.php
dougan88: @looking-4-justice I've read a couple of times that this is a good practice to use gc if you use Mockery. But can delete if you want.

c0d39uru: Every single line we add to our code/tests adds to the maintenance cost. I, personally, have never seen using this to be a good practice (unless in very rare situations). 
However, if you're comfortable, please put your reference here, so that we all learn something new.

dougan88: Here's the example of this function being used:
https://spiral-framework.com/guide/application-testing
Here's another article, but it's in Ukrainian. Listings are in php though.
https://dou.ua/lenta/articles/php-unit-testing/
Although, once again. I don't want this line to increase the maintenance cost, so I can delete it.

c0d39uru: Thanks for sharing the links. Honestly, I barely can see any explanation as to why this is required/preferred. Perhaps a matter of the writers' taste. So, please remove it, especially because PHP takes care of these things automatically and gc is on by default.

test/ContentHubClientTest.php
dougan88: done

test/ContentHubClientTest.php
dougan88: done

test/ContentHubClientTest.php
dougan88: done

test/ContentHubClientTest.php
c0d39uru: These constants are already defined for you in Response class (The Symfony one not that of Guzzle)

test/ClientCDFObjectTest.php
c0d39uru: I still can see try/catch in your tests.

dougan88: fixed

test/ContentHubClientTest.php
dougan88: fixed

c0d39uru (approved):

PR #74, opened 2019-04-30
[DRAFT] LCH-2821 Drupal should track requests from plexus

Max-Yudin:


src/ContentHubClient.php
```diff
@@ -1157,4 +1159,20 @@ protected static function removeAllLeadingAndTrailingSlashes(array $components):
             return trim($component, '/');
         }, $components);
     }
+
+    /**
+     * @param array $config
+     */
+    protected function enableRequestLogging(array $config): void
```
c0d39uru: Seems like enableRequestLogging is a misnomer. As the code indicates, it does not enable logging, but instead it adds a request logger handler to the stack.

src/Guzzle/MiddleWare/RequestLogger.php
```diff
@@ -0,0 +1,141 @@
...
+        };
+    }
+
+    /**
+     * Logs response.
+     *
+     * @param \GuzzleHttp\Psr7\Request $request
+     * @param \Psr\Http\Message\ResponseInterface $response
+     */
+    public function logResponse(Request $request, ResponseInterface $response): void {
```
c0d39uru: Can we use an interface for the first argument, too?

src/Guzzle/MiddleWare/RequestLogger.php
```diff
@@ -0,0 +1,141 @@
...
+        $entry = implode('. ', $this->buildLogEntry($request, $response));
+
+        $responseStatusCode = $response->getStatusCode();
+        switch ($responseStatusCode) {
+            case $responseStatusCode >= 500:
+                $this->logger->error($entry);
+                break;
+            case $responseStatusCode >= 400:
+                $this->logger->warning($entry);
+                break;
```
c0d39uru: 1- Use constants defined in Response class for 400 and 500
2- Switch statement is mainly for equality checks, not range check. Please use if/else instead.

src/Guzzle/MiddleWare/RequestLogger.php
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    /**
+     * @param $request
+     * @param \Psr\Http\Message\ResponseInterface $response
+     *
+     * @return array
+     */
+    protected function buildLogEntry($request, ResponseInterface $response)
```
c0d39uru: Why there's no typehint for the first param?

src/Guzzle/MiddleWare/RequestLogger.php
```diff
@@ -0,0 +1,141 @@
...
+        $messages = [];
+        $messages[] = sprintf('Request ID: %s', $body['request_id']);
+        $messages[] = sprintf(
+          'Method: %s. Path: %s. Status code: %d',
+          $request->getMethod(),
+          $request->getUri()->getPath(),
+          $response->getStatusCode()
+        );
+        $messages[] = sprintf('Body: %s', $response->getBody());
+
```
c0d39uru: Array initialization is way less expensive than insertion. The optimized way is this:
``` php
return [
    sprintf('Request ID: %s', $body['request_id']),
    sprintf(
          'Method: %s. Path: %s. Status code: %d',
          $request->getMethod(),
          $request->getUri()->getPath(),
          $response->getStatusCode()
        ),
     sprintf('Body: %s', $response->getBody()), 
];

File: src/Guzzle/MiddleWare/RequestLogger.php [hunk 6]
```diff
@@ -0,0 +1,141 @@
...
+    protected $logger;
+
+    /**
+     * RequestLogger constructor.
+     *
+     * @param \Psr\Log\LoggerInterface $logger
+     */
+    public function __construct(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
```
c0d39uru: Majority of methods of this class accept Request and/or Response as their argument(s). This means that Request and Response should be considered as a class member instead (especially that your class name is RequestLogger but surprisingly it's not dependent on Request). This also eliminates the decodeResponse method, since you can do it once and for all.

src/ContentHubClient.php
Max-Yudin: Renamed

src/Guzzle/MiddleWare/RequestLogger.php
Max-Yudin: Refactored.

src/Guzzle/MiddleWare/RequestLogger.php
Max-Yudin: Refactored.

src/Guzzle/MiddleWare/RequestLogger.php
Max-Yudin: Refactored.

src/Guzzle/MiddleWare/RequestLogger.php
Max-Yudin: Refactored.

Max-Yudin:
Hi @looking-4-justice 

All comments was processed. Please review again.

src/Guzzle/MiddleWare/RequestLogger.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    protected $decodedResponseBody;
+
+    /**
+     * RequestLogger constructor.
+     *
+     * @param \Psr\Log\LoggerInterface $logger
+     */
+    public function __construct(LoggerInterface $logger)
+    {
```
c0d39uru: If you accept that Request and Response are inevitable parts of the class, then I expect to see them as constructor arguments as well. 
On a different note, decodedResponseBody could be defined private since it's a computed class member.

Max-Yudin: Request and Response objects only available in promise function (61 line). We can't pass it into constructor.

c0d39uru: In that case, this class has more than one responsibility and should be broken down. I would rethink the design.

Max-Yudin: Logic related to logging moved to separate class.

src/Guzzle/MiddleWare/RequestResponseLogger.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        // Dont't track requests without ID.
+        if (empty($this->decodedResponseBody['request_id'])) {
+            $message = sprintf(
+                'Could not log response without ID. Response body: %s',
+                $this->response->getBody()
+            );
+            throw new Exception($message);
+        }
+
```
c0d39uru, line 77: It's not isTrackable's responsibility to throw an exception if request_id is empty. If a response with no request_id is not trackable, simply return false. If you need to throw an exception as well, move it to where it belongs!
Everything else looks good to me.

Max-Yudin: Done. Thanks for review.

c0d39uru (approved):

PR #76, opened 2019-05-15
LCH-2862 | Adding parameters to updateWebhook to support change of HMAC version and disable_retries

abarriosr:


c0d39uru (approved):

src/ContentHubClient.php
```diff
@@ -674,17 +674,27 @@ public function deleteWebhook($uuid)
...
      *
      * @throws \GuzzleHttp\Exception\RequestException
      */
-    public function updateWebhook($uuid, $url)
+    public function updateWebhook($uuid, $url, $hmac_version = 2, $disable_retries = NULL)
     {
-        $options['body'] = json_encode(['url' => $url]);
-
+        $data = ['url' => $url];
+        if (in_array($hmac_version, [1, 2])) {
```
saxumVermes: Please use the third parameter of in_array().

src/ContentHubClient.php
```diff
@@ -676,18 +676,22 @@ public function deleteWebhook($uuid)
...
+     *   The UUID of the webhook to update.
+     * @param array $options
+     *   What to change in the webhook: url, version, disable_retries, etc.
      *
      * @return \Psr\Http\Message\ResponseInterface
      *
      * @throws \GuzzleHttp\Exception\RequestException
      */
-    public function updateWebhook($uuid, $url)
+    public function updateWebhook($uuid, array $options)
```
EclipseGc, line 688: Let's get a list of "acceptable" values that could be in $options so that we don't just put anything. Loop over the acceptable value keys and only use their values.

EclipseGc (changes requested):

EclipseGc (approved):

PR #78, opened 2019-05-29
LCH 2908 | Get Interests method

kaynenh:
Added a method for getting interests by webhook

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -632,7 +632,7 @@ class ContentHubClient extends Client
 
         $data = $this->getResponseJson($this->get("/interest/webhook/$webhook_uuid"));
 
-        return !empty($data['data']['interests']) ? $data['data']['interests'] : [];
+        return $data['data']['interests'] ?? [];
 
     }
 
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -629,9 +629,11 @@ class ContentHubClient extends Client
      */
     public function getInterestsByWebhook($webhook_uuid)
     {
+
         $data = $this->getResponseJson($this->get("/interest/webhook/$webhook_uuid"));
 
-        return $data['data']['interests'];
+        return !empty($data['data']['interests']) ? $data['data']['interests'] : [];
+
     }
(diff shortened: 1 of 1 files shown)


src/ContentHubClient.php
```diff
@@ -620,6 +620,20 @@ public function getWebHook($url)
...
+     *
+     * @return array
+     *
+     * @throws \Exception
+     */
+    public function getInterestsByWebhook($webhook_uuid)
+    {
+        $data = $this->getResponseJson($this->get("/interest/webhook/$webhook_uuid"));
+
+        return $data['data']['interests'];
```
c0d39uru: What happens if:
1- `$webhook_uuid` is not valid?
2- `$webhook_uuid` does not exist?
3-  $data['data'] has not `interests` set?
4- `$this->get("/interest/webhook/$webhook_uuid")` is not a valid JSON?

kaynenh: I've added a try...catch to getResponseJson() and getInterestsByWebhook() will return an empty array now too.

src/ContentHubClient.php
```diff
@@ -629,9 +629,11 @@ public function getWebHook($url)
      */
     public function getInterestsByWebhook($webhook_uuid)
     {
+
         $data = $this->getResponseJson($this->get("/interest/webhook/$webhook_uuid"));
 
-        return $data['data']['interests'];
+        return !empty($data['data']['interests']) ? $data['data']['interests'] : [];
```
```php
        return $data['data']['interests'] ?? [];
```

kaynenh: Yes, good point.

src/ContentHubClient.php
c0d39uru: Did you check if this was necessary at all by testing the current behavior? Because chances are that these have been already dealt with.

kaynenh: I did look at it and it should at minimum cover for any issues with the plexus request... let's say if webhook_uuid is an empty string, etc.... or if something changes unknowingly with the api.

c0d39uru (approved):

PR #79, opened 2019-06-04
LCH-2395: Search criteria object [added]

Max-Yudin:


src/ContentHubClient.php
```diff
@@ -1199,4 +1202,40 @@ protected function addRequestResponseHandler(array $config): void
...
         $config['handler']->push(new RequestResponseHandler($this->logger));
     }
+
+    /**
+     * Appends search criteria header.
+     *
+     * @param array $args
+     *   Method arguments.
+     *
+     * @return mixed
```
jabxjab: @Max-Yudin Please add a return description

src/ContentHubClient.php
```diff
@@ -1199,4 +1202,40 @@ protected function addRequestResponseHandler(array $config): void
...
+            return $args;
+        }
+
+        list($baseUri, $queryString) = explode('?', $uri);
+        if (empty($queryString)) {
+            return $args;
+        }
+
+        parse_str($queryString, $parsedQueryString);
+        $messageBody = isset($options['body']) ? $options['body'] : '';
```
jabxjab: @Max-Yudin `$mesageBody` is not used

src/ContentHubClient.php
```diff
@@ -1199,4 +1202,40 @@ protected function addRequestResponseHandler(array $config): void
...
+     * @param array $args
+     *   Method arguments.
+     *
+     * @return mixed
+     */
+    protected function addSearchCriteriaHeader($args)
+    {
+        $uri = isset($args[0]) ? $args[0] : '';
+        $options = isset($args[1]) ? $args[1] : [];
+
```
c0d39uru: Use null coalesce operator

src/ContentHubClient.php
c0d39uru: Use null coalesce operator here, too

src/ContentHubClient.php
```diff
@@ -1199,4 +1202,40 @@ protected function addRequestResponseHandler(array $config): void
...
+        list($baseUri, $queryString) = explode('?', $uri);
+        if (empty($queryString)) {
+            return $args;
+        }
+
+        parse_str($queryString, $parsedQueryString);
+        $messageBody = isset($options['body']) ? $options['body'] : '';
+
+        $criteria = SearchCriteriaBuilder::createFromArray($parsedQueryString);
+        $criteria = json_encode($criteria);
```
c0d39uru: I would merge these two into one, since the result of the first assignment is only used in the  immediate statement.

src/ContentHubClient.php
```diff
@@ -1199,4 +1202,40 @@ protected function addRequestResponseHandler(array $config): void
...
+        //$criteria = base64_encode($criteria);
+        if(!isset($args[1]['headers'])){
+            $args[1]['headers'] = [];
+        }
+
```
c0d39uru, line 1237:
```php
$args[1]['headers'] = $args[1]['headers'] ?? [];
```

c0d39uru (approved):

src/SearchCriteria/SearchCriteria.php
```diff
@@ -0,0 +1,216 @@
...
+use DateTimeInterface;
+
+/**
+ * Class SearchCriteria
+ *
+ * @package Drupal\acquia_contenthub\Client
+ */
+class SearchCriteria implements \JsonSerializable
+{
+    const HEADER_NAME = 'X-Search-Criteria';
```
ynx: "X-Search-Criteria" is quite generic. Should we prefix this? What else is in the request headers right now? Any precedence we can follow, say naming it "X-Acquia-Content-Hub-Search-Criteria"? I am not married to the idea of any actual wording like "Acquia-Content-Hub", just thought it might be problem-prone if too generic.

src/SearchCriteria/SearchCriteriaBuilder.php
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @package Drupal\acquia_contenthub\Client
+ */
+class SearchCriteriaBuilder
+{
+
+    /**
+     * SearchCriteriaBuilder constructor.
+     */
+    protected function __construct()
```
saxumVermes, line 18: Is there any reason for defining constructor if it is not used? Looks like it is to prevent instantiation, what about abstract class?

PR #80, opened 2019-06-05
LCH-2940: Ensure the data payload exists before returning it.

japerry:


c0d39uru (approved):

kaynenh:
The changes in this PR must have been rolled into another commit at some point. See https://github.com/acquia/content-hub-php/blob/2.x/src/ContentHubClient.php#L652

PR #87, opened 2019-09-11
LCH-3331 | Only count on countable values for php7.2

matthewmorley:


kaynenh (approved):
Looks good to me. Not sure if there's any better way to do this.

PR #91, opened 2019-10-21
LCH-3313: Change CDFDocument to containt two maps of dependencies and ancestors; change  adding and removing CDFEntity methods to keep the new maps up-to-date

c0d39uru:
…; change  adding and removing CDFEntity methods to keep the new maps up-to-date


kaynenh:
See https://github.com/acquia/content-hub-php/pull/90

PR #93, opened 2019-11-11
LCH-3573 | enable coveralls for 1.3.x

matthewmorley:


kaynenh:
I believe this was also already done. Please let me know if that's also not right.

PR #94, opened 2019-11-11
LCH-3579 | enable coveralls for 1.2.x

matthewmorley:


kaynenh:
I think this was done already, but please re-open if I'm wrong.

PR #96, opened 2019-12-17
LCH-3642: Code style fixes.

Max-Yudin:


src/ContentHubClient.php
```diff
@@ -1056,7 +1056,7 @@ public function __call($method, $args)
     {
         try {
             if (strpos($args[0], '?')) {
-                [$uri, $query] = explode('?', $args[0]);
```
c0d39uru, line 975: Do we have PHP version issue with this?

Max-Yudin: Yes, minimal PHP version is 5.6
https://github.com/acquia/content-hub-php/blob/2.x/composer.json#L17

ynx (approved):
A lot of rules here are not Drupalism rules. I have no problem with them but I will defer to @EclipseGc to set our standards here. @EclipseGc please review.

jose-oliveira (approved):

c0d39uru (approved):

PR #97, opened 2019-12-18
LCH-3641: Add unit test to the Library

c0d39uru:
- Reviewing existing tests
  - Correcting those that are qualified to keep
  - Otherwise, rewriting them from scratch
- Adding new tests to cover the remaining parts
- Adding `infection` to gage the quality of the tests

If you want to run infection, you should install it first through `make install` command. For more information, please view the README.md file

Original ticket: https://backlog.acquia.com/browse/LCH-3641


kaynenh:
Initial overview of this looks good. Are we okay with the infection library not working on php 7.1 (failing tests error)? I believe @EclipseGc and @abarriosr may have additional thoughts around the code done here.

c0d39uru:
> Initial overview of this looks good. Are we okay with the infection library not working on php 7.1 (failing tests error)? I believe @EclipseGc and @abarriosr may have additional thoughts around the code done here.

You're right. As per Infection's composer.json, it needs `"php": "^7.2.9",` to run, but this is a dev requirement, not an actual one. So, there's no harm in letting it sit there.

composer.json
```diff
@@ -29,7 +30,9 @@
         "phpunit/phpunit": "^7",
         "scrutinizer/ocular": "~1.0",
         "sebastian/phpcpd": "~2.0",
-        "mockery/mockery": "^1.2"
+        "mockery/mockery": "^1.2",
+        "infection/infection": "^0.15.0",
+        "ext-json": "*"
```
jose-oliveira: This dependency is added both as prod and dev.

jose-oliveira:
> > Initial overview of this looks good. Are we okay with the infection library not working on php 7.1 (failing tests error)? I believe @EclipseGc and @abarriosr may have additional thoughts around the code done here.
> 
> You're right. As per Infection's composer.json, it needs `"php": "^7.2.9",` to run, but this is a dev requirement, not an actual one. So, there's no harm in letting it sit there.

Well, the harm is that we can't run tests on 7.1 with this dependency. If we aren't running Infection as part of our CI process, maybe we should require it in the make file?

jose-oliveira (changes requested):
Code looks ok, but we need to figure out what to do with PHP 7.1 tests failing due to incompatibility with Infection version.

composer.json
c0d39uru: @jose-oliveira Mockery has been there for a while, I didn't add it.

cleancode4ever:
> Code looks ok, but we need to figure out what to do with PHP 7.1 tests failing due to incompatibility with Infection version.

That make

c0d39uru:
That's right. I think we can safely remove infection for a while.

composer.json
jose-oliveira: I meant the ext-json one

c0d39uru: Fixed

jose-oliveira (approved):

src/ContentHubClient.php
```diff
@@ -305,17 +309,16 @@ public function getEntities(array $uuids)
                 ],
               ],
             ];
-            $options['body'] = json_encode((array)$query);
-            $results = $this->getResponseJson($this->get('_search', $options));
-            if (is_array($results) && isset($results['hits']['total']) && $results['hits']['total'] > 0) {
+            $options['body'] = json_encode($query);
+            $results = self::getResponseJson($this->get('_search', $options));
+            if (isset($results['hits']['total'])) {
```
EclipseGc, line 314: In general I prefer !empty() to isset(). We can do that in a follow up though.

src/SearchCriteria/SearchCriteriaBuilder.php
```diff
@@ -80,11 +72,8 @@ protected static function propertiesMap()
     protected static function extractPropertyFromArray(string $propertyName, array $data, $default = '')
     {
         $map = self::propertiesMap();
-        if (!isset($map[$propertyName])) {
```
EclipseGc, line 83: Discussed the removal of this bit. @cleancode4ever pointed out that it was overly paranoid, and that with the way the system is formulated he couldn't actually test it meaning there wasn't effectively a way to ever invoke it.

src/ContentHubClient.php
```diff
@@ -328,51 +331,12 @@ public function getEntities(array $uuids)
...
-  /**
-   * Updates many entities asynchronously.
-   *
-   * @param \Acquia\ContentHubClient\CDF\CDFObject[] $objects
-   *   The CDFObjects to update.
-   *
-   * @return \Psr\Http\Message\ResponseInterface
-   *
-   */
-    public function putEntities(CDFObject ...$objects)
```
EclipseGc, line 363: \Drupal\acquia_contenthub_publisher\Plugin\QueueWorker Line 171 actually does a putEntities(). I think it's postEntities we don't use.

EclipseGc (changes requested):
Removing putEntities isn't going to work for us. It's postEntities() we never use. (also we don't use putEntity)

src/ContentHubClient.php
c0d39uru: I added it back!

src/ContentHubClient.php
c0d39uru: There are looooooooooots of other places that need such changes. Now, with having so many tests in place, it'll be like a breeze to refactor and make it better.

c0d39uru:
> Removing putEntities isn't going to work for us. It's postEntities() we never use. (also we don't use putEntity)

Fixed!

PR #98, opened 2020-02-10
LCH-3773: The response body removed from the log entry structure.

Max-Yudin:


c0d39uru (approved):

src/Guzzle/Middleware/RequestResponseLogger.php
```diff
@@ -92,12 +97,11 @@ protected function isTrackable(): bool
...
     {
         return sprintf(
-            'Request ID: %s. Method: %s. Path: %s. Status code: %d. Body: %s',
+            'Request ID: %s. Method: %s. Path: %s. Status code: %d.',
             $this->decodedResponseBody['request_id'],
             $this->request->getMethod(),
             $this->request->getUri()->getPath(),
-            $this->response->getStatusCode(),
-            $this->response->getBody()
+            $this->response->getStatusCode()
```
kaynenh, line 104: I do wonder whether we should actually leave the body if we are electing to just not track anything with a shared secret in it. What you've done fixes the immediate problem, but we may want to look at this again in the future if we were to log the body again.

kaynenh (approved):
This looks good Max. Thanks!

src/Guzzle/Middleware/RequestResponseLogger.php
EclipseGc: I'm ok with how this is for the time being. If we decide we need to change what's logged later, we can tackle that issue at that time.

PR #101, opened 2020-03-02
LCH-3485: Add functions to handle webhook suppression.

Ateee89:


src/ContentHubClient.php
```diff
@@ -787,6 +787,30 @@ public function updateWebhook($uuid, array $options) {
...
+   * Remove suppression from webhook.
+   *
+   * @param string $webhook_uuid
+   *   Webhook uuid.
+   *
+   * @return \Psr\Http\Message\ResponseInterface
+   */
+  public function unSuppressWebhook(string $webhook_uuid) {
+    return $this->delete("webhook/$webhook_uuid/suppress");
+  }
```
c0d39uru: For the record, this is not the right way of interacting with an API as a client library. Client library (in this case, Content Hub PHP, is supposed to not only talk to the API, but also take care of all the possible results including success and failures. 
I know there are other functions in this library that does the exact same thing (simply returning the raw http call), but it's legacy. Now that we are adding a new one, we should do it the right way.

Keep in mind, the library's consumer should 100% rely on the library for error checking and handling. The purpose of he library is not to make the HTTP call only.

c0d39uru (approved):

c0d39uru:
@Ateee89 Any idea why coveralls is reporting coverage decrease?

EclipseGc:
Can we get test coverage for these two new methods?

Ateee89:
@EclipseGc @cleancode4ever Sorry totally forgot about the coverage, fixed now.

PR #102, opened 2020-03-16
LCH-3571: (2.x) Return DEPRECATED when logs method gets called

c0d39uru:
As part of retiring logs history feature from Plexus, this PR aims at returning DEPRECATED response when logs method gets called instead of hitting Plexus.

Subtask: https://backlog.acquia.com/browse/LCH-3571

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -39,11 +39,11 @@ class ContentHubClient extends Client {
 
   const OPTION_NAME_LANGUAGES = 'client-languages';
 
-  const FEATURE_DISCONTINUED_RESPONSE = [
+  const FEATURE_DEPRECATED_RESPONSE = [
       'success' => FALSE,
       'error' => [
         'code' => SymfonyResponse::HTTP_GONE,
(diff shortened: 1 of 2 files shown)


src/ContentHubClient.php
```diff
@@ -38,6 +39,14 @@ class ContentHubClient extends Client {
 
   const OPTION_NAME_LANGUAGES = 'client-languages';
 
+  const FEATURE_DISCONTINUED_RESPONSE = [
+      'success' => FALSE,
+      'error' => [
+        'code' => SymfonyResponse::HTTP_GONE,
+        'message' => 'This feature is discontinued',
```
kaynenh: Maybe "This feature is deprecated"?

kaynenh (approved):
This makes sense to me. Possibly change discontinued to deprecated, but that's my only thought.

src/ContentHubClient.php
c0d39uru: @kaynenh Sure, if that makes more sense, I'll change it.

PR #109, opened 2020-04-01
LCH-2722 | Improve error handling in the php library

abarriosr:


src/ContentHubClient.php
```diff
@@ -1116,151 +1117,111 @@ public function __call($method, $args) {
    * @param \Exception $exception
    *   The Exception object.
    *
-   * @return \Psr\Http\Message\ResponseInterface
-   *   The text to write in the messages.
-   *
-   * @codeCoverageIgnore
```
kaynenh, line 1122: We used to ignore this... we probably still should.

kaynenh (approved):
I believe this looks good now. We should follow-up on why we are ignoring coveralls for some of these methods in content-hub-php.

PR #113, opened 2020-06-15
LCH-4047: Fix endpoint call in webhook suppression method.

Ateee89:


EclipseGc:
This seems super obvious. Merging.

PR #116, opened 2020-10-26
LCH-4486-1.x: Ability to take snapshots from plexus

ankitsingh0188:


src/ContentHub.php
```diff
@@ -583,4 +583,62 @@ protected function getResponseJson(RequestInterface $request)
...
+     * Fetch snapshots.
+     *
+     * @return mixed
+     *   Response.
+     *
+     * @throws \Exception
+     */
+    public function getSnapshots()
+    {
+      return self::getResponseJson($this->get('snapshots'));
```
Ateee89: I am not sure how can this work. getResponseJson() function first parameter should be a RequestInterface. $this->get('snapshots') will return a ResponseInterface

src/ContentHub.php
```diff
@@ -583,4 +583,62 @@ protected function getResponseJson(RequestInterface $request)
...
+     * Create a snapshot.
+     *
+     * @return mixed
+     *   Response.
+     *
+     * @throws \Exception
+     */
+    public function createSnapshot()
+    {
+      return self::getResponseJson($this->post('snapshots'));
```
Ateee89: Same here

src/ContentHub.php
```diff
@@ -583,4 +583,62 @@ protected function getResponseJson(RequestInterface $request)
...
+     *   The name of the snapshot.
+     *
+     * @return \Psr\Http\Message\ResponseInterface
+     *   Response.
+     *
+     * @throws \GuzzleHttp\Exception\RequestException
+     */
+    public function deleteSnapshot($name)
+    {
+      return self::getResponseJson($this->delete("snapshots/$name"));
```
Ateee89: Same here

src/ContentHub.php
```diff
@@ -583,4 +583,62 @@ protected function getResponseJson(RequestInterface $request)
...
+     *   The name of the snapshot.
+     *
+     * @return \Psr\Http\Message\ResponseInterface
+     *   Response.
+     *
+     * @throws \GuzzleHttp\Exception\RequestException
+     */
+    public function restoreSnapshot($name)
+    {
+      return self::getResponseJson($this->put("snapshots/$name/restore"));
```
Ateee89: Same here.

Ateee89 (changes requested):

src/ContentHub.php
```diff
@@ -583,4 +583,70 @@ protected function getResponseJson(RequestInterface $request)
...
+     * Fetch snapshots.
+     *
+     * @return mixed
+     *   Response.
+     *
+     * @throws \Exception
+     */
+    public function getSnapshots()
+    {
+      $endpoint = "$this->api_version/snapshots";
```
abarriosr: The only available version for this API call is `/v2`. Then all calls should use that instead of `$api_version`.

Ateee89 (approved):

Ateee89 (approved):

PR #119, opened 2021-01-03
D.O. #3190725, Allow to update the dependency hash in later phase.

weynhamz:
https://www.drupal.org/project/acquia_contenthub/issues/3190725


amangrover90:
Not needed anymore as hash is same in both the places now.

PR #122, opened 2021-07-22
Update composer to support only Drupal 8.9+

japerry:
We should tag this as version 1.4.x


japerry:
Merging into 1.3.x because we do not support the older versions of phpunit or PHP 7/5.6.

PR #124, opened 2021-08-24
LCH-5312: Created filter functions and code refactor

ankitsingh0188:


src/ContentHubClient.php
```diff
@@ -1182,13 +1183,13 @@ protected function getExceptionResponse($method, array $args, \Exception $except
...
    *   Reason.
-   * @param null $request_id
+   * @param string|null $request_id
    *   The request id from the ContentHub service if available.
    *
    * @return \GuzzleHttp\Psr7\Response
    *   Response.
    */
-  protected function getErrorResponse($code, $reason, $request_id = NULL) {
+  protected function getErrorResponse(int $code, string $reason, ?string $request_id = NULL): Response {
```
saxumVermes: Use response interface instead as the return type.

src/ContentHubClient.php
```diff
@@ -1359,13 +1360,85 @@ public function deleteSnapshot($name) {
...
+   *   suffixed with duration unit (m, s, ms etc.).
+   * @param int $size
+   *   Amount of entities to return.
+   *
+   * @return mixed
+   *   Response from backend call.
+   *
+   * @throws \Exception
+   */
+  public function startScrollByFilter(string $filter_uuid, string $scroll_time_window, int $size) {
```
saxumVermes: So although the json_decode can return objects too, we are actually command it to return an associative array. Please specify the return type accordingly.

src/ContentHubClient.php
```diff
@@ -1359,13 +1360,85 @@ public function deleteSnapshot($name) {
...
+   *   Scroll id.
+   * @param string $scroll_time_window
+   *   How long the scroll cursor will be retained inside memory.
+   *
+   * @return mixed
+   *   Response from backend call.
+   *
+   * @throws \Exception
+   */
+  public function continueScroll(string $scroll_id, string $scroll_time_window) {
```
saxumVermes: Same here about return types.

src/ContentHubClient.php
```diff
@@ -1359,13 +1360,85 @@ public function deleteSnapshot($name) {
...
+   *
+   * @param string $scroll_id
+   *   Scroll id.
+   *
+   * @return mixed
+   *   Response from backend call.
+   *
+   * @throws \Exception
+   */
+  public function cancelScroll(string $scroll_id) {
```
saxumVermes: And here.

src/ContentHubClient.php
```diff
@@ -1359,13 +1360,85 @@ public function deleteSnapshot($name) {
...
   }
 
+  /**
+   * Initiates Scroll API request chain.
+   *
+   * @param string $filter_uuid
+   *   Filter uuid to execute by.
+   * @param string $scroll_time_window
+   *   How long the scroll cursor will be retained inside memory. Must be
+   *   suffixed with duration unit (m, s, ms etc.).
```
saxumVermes, line 1378: Hope we can get a response about that soon...

test/ContentHubClientTest.php
```diff
@@ -2412,4 +2412,92 @@ public function testRestoreSnapshotIfSucceeds(): void {
...
+   *
+   * @throws \Exception
+   */
+  public function testContinueScrollIfSucceeds(): void {
+    $scroll_id = 'some-scroll-id';
+    $scroll_time_window = '10m';
+
+    $request_parameters = [
+      'scroll_id' => $scroll_id,
+      'scroll' => '10m',
```
saxumVermes: You can just use `$scroll_time_window` here too, can't you?

test/ContentHubClientTest.php
```diff
@@ -2412,4 +2412,92 @@ public function testRestoreSnapshotIfSucceeds(): void {
...
+    $response = [
+      'uuid' => 'some-uuid',
+      'request_id' => 'some-request-id',
+      'success' => TRUE,
+    ];
+
+    $this->ch_client
+      ->shouldReceive('delete')
+      ->once()
+      ->with("scroll", ['body' => json_encode($request_parameters)])
```
saxumVermes: Could you please use single quotes?

saxumVermes (changes requested):

src/ContentHubClient.php
ankitsingh0188: Yeah, awaiting response from service team.

src/ContentHubClient.php
```diff
@@ -3,27 +3,23 @@
...
 use GuzzleHttp\Exception\BadResponseException;
 use GuzzleHttp\Exception\ClientException;
-use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Exception\RequestException;
-use GuzzleHttp\Exception\ServerException;
 use GuzzleHttp\HandlerStack;
 use GuzzleHttp\Psr7\Response;
 use Psr\Log\LogLevel;
+use Symfony\Component\HttpFoundation\Response as HttpResponse;
 use Symfony\Component\HttpFoundation\Response as SymfonyResponse;
```
saxumVermes: I don't understand that, why is that we're importing the same object twice with different names?

saxumVermes (approved):
Left a comment, there's an unnecessary import, please fix that, otherwise it's good to go.

src/ContentHubClient.php
ankitsingh0188: Done.

amangrover90 (approved):
Looks good to me!!

PR #141, opened 2022-03-16
http-hmac-php v6

marthinal:
We have some conflicts upgrading to PHP 8

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "issues": "https://github.com/acquia/content-hub-php/issues"
     },
     "require": {
-        "acquia/http-hmac-php": "~3.4|~4.1.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
         "issues": "https://github.com/acquia/content-hub-php/issues"
     },
     "require": {
-        "acquia/http-hmac-php": "~3.4|~4.1.2",
```
japerry, line 17: We really only need to specify a minimum here. The http-hmac-php spec isn't going to be making future changes that this library won't be compatible with.

I'd suggest instead just changing it to `>=3.4`

japerry (changes requested):

amangrover90:
This has already been fixed in #142

PR #148, opened 2022-06-03
LCH-6061: Fix deprecations for symfony 3.4.

amangrover90:


saxumVermes (changes requested):
I think here as well, we can revise the dispatch() arguments. From Symfony 5 ordering changes, better get ahead of that deprecation.

saxumVermes:
Also all the symfony 3.4 from composer.json needs to be removed.

amangrover90:
> 

Yeah I missed that, was gonna do it after lunch.

narendradesai (approved):
LGTM 👍🏻

saxumVermes (approved):
Looks good! Let's roll it!

PR #149, opened 2022-06-09
LCH-6084: Fixing php8 deprecations.

narendradesai:


src/ContentHubClient.php
```diff
@@ -85,11 +85,11 @@ class ContentHubClient extends Client {
    * @codeCoverageIgnore
    */
   public function __construct(
-    array $config = [],
     LoggerInterface $logger,
     Settings $settings,
     HmacAuthMiddleware $middleware,
     EventDispatcherInterface $dispatcher,
+    array $config = [],
```
amangrover90, line 92: Out of scope of this ticket: We shouldn't be using inheritdoc for function doc, as it has different arguments from its parent class.

amangrover90 (approved):
LGTM!!

saxumVermes (approved):

PR #150, opened 2022-06-27
LCH-6094: Change logging client to send multiple logs at once

amangrover90:


src/ContentHubLoggingClient.php
```diff
@@ -19,7 +19,7 @@ class ContentHubLoggingClient extends Client {
   use ContentHubClientTrait;
 
   // Override VERSION inherited from GuzzleHttp::ClientInterface.
-  const VERSION = '2.0.0';
+  const LIB_VERSION = '2.0.0';
```
saxumVermes: Hmm, wondering if we need that. The most appropriate solution would be to move the lib version and the user agent string to a separate, final class. Hence we'd only need to modify it at one place and we won't forget to change the version here too.
Could you please do that? We can call the class ContentHubDescriptor final class.

src/ContentHubLoggingClient.php
```diff
@@ -71,7 +71,7 @@ public function __construct(
     }
 
     // Setting up the User Header string.
-    $user_agent_string = self::LIBRARYNAME . '/' . self::VERSION . ' ' . default_user_agent();
+    $user_agent_string = self::LIBRARYNAME . '/' . self::LIB_VERSION . ' ' . default_user_agent();
```
saxumVermes: This would become `ContentHubDescriptor::userAgent()`

saxumVermes (changes requested):

saxumVermes (approved):
Looks great, let's roll it!

PR #154, opened 2022-10-11
LCH-6162: D10 readiness deprecations.

amangrover90:


narendradesai (changes requested):
~~Changes looks good. However upgrade status tool still shows 100 errors, although those are related to tests, drupal canary is still going to say module not compatible with D10.
[single-export-acquia_contenthub-2022-10-12T13_11_58+0000.txt](https://github.com/acquia/content-hub-php/files/9765185/single-export-acquia_contenthub-2022-10-12T13_11_58%2B0000.txt)~~

This was for : https://github.com/acquia/acquia_contenthub/pull/879

saxumVermes:
I extended the test run to php8.1 as well, however it started failing due to a class name  `Mixed` which is reserved since the php8.0. I changed it to MixedStandardizer.

japerry (approved):

PR #157, opened 2022-12-15
LCH-6253: Fix php-stan issues in content-hub-php

narendradesai:
Issue: https://backlog.acquia.com/browse/LCH-6253

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "pdepend/pdepend": "~1.0",
         "phploc/phploc": "~2.0",
         "phpmd/phpmd": "~1.0",
-        "phpunit/phpunit": "^8",
+        "phpunit/phpunit": "^9",
         "scrutinizer/ocular": "~1.0",
         "sebastian/phpcpd": "~2.0",
         "squizlabs/php_codesniffer": "^3.5"
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "ext-json": "*",
         "php": ">=7.3",
         "psr/log": ">=1.0",
-        "guzzlehttp/guzzle": ">=6.5 <=7.5.0",
+        "guzzlehttp/guzzle": ">=6.5 <8.0",
         "symfony/event-dispatcher": ">=4.4",
         "symfony/http-foundation": ">=4.4",
         "symfony/serializer": ">=4.4"
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,9 @@
         "phpunit/phpunit": "^9",
         "scrutinizer/ocular": "~1.0",
         "sebastian/phpcpd": "~2.0",
-        "squizlabs/php_codesniffer": "^3.5"
+        "squizlabs/php_codesniffer": "^3.5",
+        "phpstan/phpstan": "^1.8",
+        "phpspec/prophecy-phpunit": "^2"
     },
     "suggest": {
         "ramsey/uuid": "A PHP 5.4+ library for generating RFC 4122 version 1, 3, 4, and 5 universally unique identifiers (UUID)"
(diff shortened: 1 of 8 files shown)


phpstan.neon
```diff
@@ -0,0 +1,9 @@
+parameters:
+  level: 1
+  paths:
+    - ./src
+  ignoreErrors:
+    - '#Class [a-zA-Z0-9\\_:\(\).]* extends [a-zA-Z0-9\\_:@\(\).]*#'
```
saxumVermes, line 6: Why were these error ignored? @narendradesai

src/CDF/CDFObject.php
```diff
@@ -235,6 +235,9 @@ public function getAttributes() {
 
   /**
    * {@inheritdoc}
+   *
+   * @return \Acquia\ContentHubClient\CDFAttribute|void
```
saxumVermes: When I'm checking the return value of the variable, in this case it would be NULL isn't it? In this case do you think it would be more appropriate to change the return type to `\Acquia\ContentHubClient\CDFAttribute|null` and specify `?CDFAttribute` as the return data type for the function?

saxumVermes (changes requested):
I left 2 minor comments, but the second requires some action, otherwise looks good!

phpstan.neon
narendradesai: @saxumVermes This added to suppress below error:
`Class Acquia\ContentHubClient\ContentHubClient extends @final class GuzzleHttp\Client`

src/CDF/CDFObject.php
narendradesai: `  /**
   * {@inheritdoc}
   *
   * @return \Acquia\ContentHubClient\CDFAttribute|void
   *   Attribute object.
   */
  public function getAttribute($id) {
    if (!empty($this->attributes[$id])) {
      return $this->attributes[$id];
    }
  }`

If the`!empty($this->attributes[$id])` condition is not satisfied then function will not return anything. Hence `void` is added.

phpstan.neon
saxumVermes: This is a really good example why this should not be done. Please next time include the reason in the solution! I've just read why they tagged the Client as final. the reason being, when version 8 comes out the will set the class (not just annotating) as final. They wanted to mark it as a deprecated that way, which will break backward compatibility with all our implementations essentially disabling - without refactor - updating to an upper version.
This needs to be taken care of soon.

src/CDF/CDFObject.php
saxumVermes: If we abide by the clean code rules, this function either should return something or nothing. There should not be in-between. If the function doesn't return an object it should return null.
```php
$value = $this->getAttribute("some_attribute");
var_dump($value);
```
The above code has two possible return types. Either a CDFAttribute object or, because a variable must have a value, NULL.

saxumVermes:
Thanks to phpstan we identified that one of the failures pointed out a future, planned deprecation: GuzzelHttp/Client will be set to final. This will prevent any implementing side to extend the class. This needs two follow-up actions:
* Create a ticket which resolves this issue to make it compatible with guzzle v8
  * We need to fix every such issue related to this deprecation and re-enable this check so future deprecations can be identified
* Set maximum version to guzzle v7 in composer.json in this ticket

@narendradesai Please create the appropriate tickets.

composer.json
```diff
@@ -18,7 +18,7 @@
         "ext-json": "*",
         "php": ">=7.3",
         "psr/log": ">=1.0",
-        "guzzlehttp/guzzle": ">=6.5",
+        "guzzlehttp/guzzle": ">=6.5 <=7.5.0",
```
saxumVermes: I think this can be `<8` right? Because according to guzzle devs this will be deprecated in version 8

saxumVermes (approved):
This ticket PR in its form is good to be merged. We'll handle the guzzle issue in a separate task.

PR #158, opened 2022-12-28
LCH-6278: New method for filter less scroll.

amangrover90:


src/ContentHubClient.php
```diff
@@ -1281,6 +1281,29 @@ public function startScrollByFilter(string $filter_uuid, string $scroll_time_win
...
+   *   suffixed with duration unit (m, s, ms etc.).
+   * @param int $size
+   *   Amount of entities to return.
+   *
+   * @return array
+   *   Response from scroll API.
+   *
+   * @throws \Exception
+   */
+  public function startScroll(string $scroll_time_window = '30m', int $size = 100): array {
```
saxumVermes: For a complete implementation we'd need to have the query part specified as well. The `/scroll` endpoint accepts an ElasticSearch compatible query under the `query ` key in the body.

saxumVermes (changes requested):

src/ContentHubClient.php
saxumVermes: Just like in case of `/_search` endpoint.

src/ContentHubClient.php
```diff
@@ -940,12 +956,15 @@ public function updateWebhook($uuid, array $options) {
    *
    * @param string $webhook_uuid
    *   Webhook uuid.
+   * @param string $suppress_duration
+   *   Duration for which webhook should be suppressed. e.g. 24h.
```
saxumVermes, line 960: Perhaps it is worth a comment if we leave the suppressed_duration blank, it will indefinitely suppress the webhook.

saxumVermes (approved):
We need test coverage for the new methods. Also I left a minor comment regarding the comment. I'll create a new ticket to address these in the next sprint. Otherwise looks good! :)

PR #159, opened 2023-01-06
LCH-6300: Use composition instead of inheritance.

saxumVermes:


src/ContentHubClientTrait.php
```diff
@@ -235,4 +250,39 @@ public function ping() {
     return $client->get('ping');
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function send(RequestInterface $request, array $options = []): ResponseInterface {
+    return $this->httpClient->send($request, $options);
```
amangrover90, line 257: Wouldn't this throw an error calling send on null? and same for other 4 methods since httpClient might not be set. Here it becomes responsibility of class using this trait to initialise httpClient.
@saxumVermes what do you think?

test/ContentHubLoggingClientTest.php
```diff
@@ -23,6 +24,8 @@
  */
 class ContentHubLoggingClientTest extends TestCase {
 
+  use ProphecyTrait;
```
narendradesai, line 27: @saxumVermes are we using this trait?

saxumVermes: Yes, this is required to avoid deprecation notices. Prophesize function is defined there.

src/ContentHubClientTrait.php
saxumVermes: That's right. The trait proposes a contract which must be satisfied by the implementing classes. ContentHubCientTrait is very specific in this case. The responsibility is delegated to the class in hand.
The most appropriate way to handle this problem is to provide the client as an argument, however that would break backward compatibility.

narendradesai (approved):
Looks good

amangrover90 (approved):
LGTM!!

PR #162, opened 2023-01-20
LCH-6321: Adding scroll endpoint.

narendradesai:


saxumVermes (changes requested):
We'd need a test to cover that one, could you please provide unit tests?

saxumVermes (approved):

PR #165, opened 2023-02-21
LCH-6356: Adding compatibility with guzzle > 7.5.  (3.x)

narendradesai:

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -12,7 +12,6 @@ use GuzzleHttp\Exception\BadResponseException;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Exception\InvalidArgumentException;
 use GuzzleHttp\Exception\RequestException;
-use GuzzleHttp\Promise\PromiseInterface;
 use GuzzleHttp\Psr7\Response;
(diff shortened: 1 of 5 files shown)


src/ContentHubClient.php
```diff
@@ -147,8 +151,16 @@ public function __call($method, $args) {
 
       $args = $this->addSearchCriteriaHeader($args);
 
-      return $this->httpClient->__call($method, $args);
+      if (\count($args) < 1) {
```
saxumVermes: Can we just remove __call? I don't think we need it at all. In related methods, such as get, put, patch, post can be passed to the guzzle http.

narendradesai:
PR for CH3.0.x module: https://github.com/acquia/acquia_contenthub/pull/960

narendradesai:
Setting minimum version of guzzle to 6.5
`drupal/core-recommended 9.4.11 requires guzzlehttp/guzzle ~6.5.8`

amangrover90 (commented):
3.x module branch's tests are failing due to method declaration incompatibility. I think the same would happen for 2.x branch.

saxumVermes (approved):
Let's roll it!

PR #166, opened 2023-02-27
LCH-6356: Adding guzzle 8 compatibility. 2.x.

narendradesai:

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -8,8 +8,11 @@ use Acquia\ContentHubClient\SearchCriteria\SearchCriteria;
 use Acquia\ContentHubClient\SearchCriteria\SearchCriteriaBuilder;
 use Acquia\Hmac\Guzzle\HmacAuthMiddleware;
 use GuzzleHttp\Client;
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\ClientTrait;
 use GuzzleHttp\Exception\BadResponseException;
 use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\InvalidArgumentException;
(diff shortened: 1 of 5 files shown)


src/ContentHubClient.php
```diff
@@ -26,7 +29,10 @@
  *
  * @package Acquia\ContentHubClient
  */
-class ContentHubClient extends Client {
+class ContentHubClient implements ClientInterface {
+
+  use ClientTrait;
```
saxumVermes: ClientTrait is not available in guzzle <7.0, however, Drupal 9.x supports guzzle 6. In this particular case the best would be to actually just implement the methods, get, put, delete, post and use the client we already have to execute those requests.

src/ContentHubClientTrait.php
```diff
@@ -0,0 +1,307 @@
...
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7\Response;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Log\LogLevel;
+
+/**
+ * Common trait for CH Client and CH Logging Client.
+ */
+trait ContentHubClientTrait {
```
saxumVermes, line 16: The implementation of get, put, post, delete can go here and we can use httpClient to carry out those requests.

narendradesai:
CH 2.x module PR: https://github.com/acquia/acquia_contenthub/pull/962

saxumVermes (approved):
Looks great!

amangrover90 (approved):
LGTM!!

saxumVermes (approved):
Looks good!

PR #167, opened 2023-03-03
LCH-6356: Fixing argument hinting.

narendradesai:

### DIFF
--- a/src/ContentHubClientTrait.php
+++ b/src/ContentHubClientTrait.php
@@ -22,7 +22,14 @@ trait ContentHubClientTrait {
    *
    * @var \GuzzleHttp\ClientInterface
    */
-  protected $httpClient;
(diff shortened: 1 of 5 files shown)


src/ContentHubClientTrait.php
```diff
@@ -152,9 +152,8 @@ protected function getErrorResponse($code, $reason, $request_id = NULL) {
    *
    *  @codeCoverageIgnore
    */
-  protected function getExceptionResponse($method, array $args, \Exception $exception) {
+  protected function getExceptionResponse($method, string $api_call, \Exception $exception) {
```
saxumVermes: The method can be more strict, it can only be a string. The return type can be marked and fixed as well.

saxumVermes (approved):
Looks good!

PR #170, opened 2023-04-05
LCH-6395: Added php 8.2 in github actions. 3.x.

narendradesai:
Issue: https://backlog.acquia.com/browse/LCH-6395


saxumVermes (approved):
Let's roll it!

PR #173, opened 2023-04-18
LCH-6426: Add log record parameter object and make logging more extensible.

saxumVermes:


src/Logging/RequestErrorLogParameter.php
```diff
@@ -0,0 +1,98 @@
+<?php
+// phpcs:ignoreFile
```
narendradesai, line 2: Any specific reason for ignoring phpcs and not adding doc comments?

saxumVermes: To avoid unnecessary boilerplate.

src/ContentHubClientTrait.php
```diff
@@ -3,6 +3,8 @@
 namespace Acquia\ContentHubClient;
 
 use Acquia\ContentHubClient\Guzzle\Middleware\RequestResponseHandler;
+use Acquia\ContentHubClient\Logging\LoggingHelperTrait;
+use Drupal\Core\Logger\LoggerChannelTrait;
```
narendradesai: Unused class

src/ContentHubClientTrait.php
```diff
@@ -3,6 +3,8 @@
...
 use GuzzleHttp\HandlerStack;
 use GuzzleHttp\Promise\PromiseInterface;
```
narendradesai: Unused PromiseInterface

src/ContentHubClientTrait.php
```diff
@@ -3,6 +3,8 @@
...
 use GuzzleHttp\Psr7\Response;
```
narendradesai: Unused

narendradesai: `LogLevel` is also not used

src/Logging/LoggingHelperTrait.php
```diff
@@ -0,0 +1,159 @@
...
+   *   Status code.
+   * @param string $reason
+   *   Reason.
+   * @param string|null $request_id
+   *   The request id from the ContentHub service if available.
+   *
+   * @return \GuzzleHttp\Psr7\Response
+   *   Response.
+   */
+  protected function getErrorResponse($code, $reason, $request_id = NULL) {
```
narendradesai: Missing return type

narendradesai: Missing parameter hinting

src/Logging/RequestErrorLogParameter.php
saxumVermes: We are using drupal standards here and this file can grow up to twice its size due to that reason, while this is only a parameter object. Drupal standard, in my opinion, has some unnecessary or unrefined restrictions which results in an antipattern of over-documentation.

narendradesai (changes requested):

src/ContentHubClientTrait.php
saxumVermes: Hmm, we're gonna need to finetune phpcs. These should have been caught.

narendradesai (approved):
Looks good

PR #174, opened 2023-04-19
LCH-6425: Method to update webhook data in settings.

narendradesai:


src/Settings.php
```diff
@@ -209,4 +209,14 @@ public function getSharedSecret() {
     return $this->sharedSecret;
   }
 
+  /**
+   * Updates webhook data.
+   *
+   * @param array $webhook
+   *   The webhook data.
+   */
+  public function updateWebhook(array $webhook): void {
```
saxumVermes, line 218: I feel like we need to search for another solution. This would violate the purpose of this class, which is being an immutable object.

saxumVermes (changes requested):

src/Settings.php
saxumVermes: One its attributes of an immutable object is to preserve state and avoid any side effects.

narendradesai:
Closing this PR as it is not necessary.

PR #180, opened 2023-07-07
LCH-6559: Throw an exception if search is not responding instead of returning empty document.

amangrover90:


saxumVermes (approved):
Approved! 🎉

narendradesai (approved):
Looks good

PR #181, opened 2023-07-12
LCH-6543: Add methods for sending client metadata while registering or updating the client.

amangrover90:


test/MetaData/ClientMetaDataTest.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace test\MetaData;
```
narendradesai:
```suggestion
namespace Acquia\ContentHubClient\test\MetaData;
```

narendradesai (changes requested):
Overall looks good.

src/MetaData/ClientMetaData.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+namespace Acquia\ContentHubClient\MetaData;
+
+/**
+ * Client metadata.
+ */
+class ClientMetaData {
```
saxumVermes, line 8: I like that!

src/MetaData/ClientMetaData.php
```diff
@@ -0,0 +1,96 @@
...
+   * @param string $client_type
+   *   Client type. e.g. drupal, pwa_app, nextjs_app etc.
+   * @param bool $is_publisher
+   *   Whether this site is a publiher or not.
+   * @param bool $is_subscriber
+   *   Whether this site is a subscriber or not.
+   * @param array $client_config
+   *   Additional config metadata.
+   */
+  public function __construct(string $client_type, bool $is_publisher, bool $is_subscriber, array $client_config) {
```
saxumVermes: Perhaps we can initialise $client_config with an empty array marking it as optional.

amangrover90: That makes sense.

src/MetaData/ClientMetaData.php
```diff
@@ -0,0 +1,96 @@
...
+    throw new \RuntimeException('All the attributes: "client_type", "is_publisher", "is_subscriber" are required.');
+  }
+
+  /**
+   * Returns metadata of the client.
+   *
+   * @return array
+   *   Client metadata.
+   */
+  public function getMetadata(): array {
```
saxumVermes: Perhaps we can rename this to `toArray`? Or `getValues`? The class is already ClientMetadata, I think it already communicates clearly, that it encapsulates the client metadata.

amangrover90: Done.

saxumVermes (approved):
Looks good, I really liked the abstraction of ClientMetadata. I had a tiny concern of it not implementing an interface to comply to the SOLID guidelines, but we can move forward with it I think.

narendradesai (approved):

PR #183, opened 2023-08-23
LCH-6635: Added error message on unable to delete client.

narendradesai:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1032,7 +1032,9 @@ class ContentHubClient implements ClientInterface {
     if (!$response) {
       throw new \Exception(sprintf("Entity with UUID = %s cannot be deleted.", $uuid));
     }
-    if ($response->getStatusCode() < 200 || $response->getStatusCode() >= 300) {
+    if ($response->getStatusCode() !== 404 &&
+      ($response->getStatusCode() < 200 || $response->getStatusCode() >= 300)
+    ) {
       throw new \Exception(sprintf("Entity with UUID = %s cannot be deleted. Message: %s", $uuid, (string) $response->getBody()));
     }
     return $this->delete("settings/client/uuid/$uuid");
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1029,8 +1029,11 @@ class ContentHubClient implements ClientInterface {
     $settings = $this->getSettings();
     $uuid = $client_uuid ?? $settings->getUuid();
     $response = $this->deleteEntity($uuid);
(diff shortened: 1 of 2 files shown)


src/ContentHubClient.php
```diff
@@ -1029,9 +1029,12 @@ public function deleteClient($client_uuid = NULL) {
     $settings = $this->getSettings();
     $uuid = $client_uuid ?? $settings->getUuid();
     $response = $this->deleteEntity($uuid);
-    if (!$response || $response->getStatusCode() < 200 || $response->getStatusCode() >= 300) {
+    if (!$response) {
       throw new \Exception(sprintf("Entity with UUID = %s cannot be deleted.", $uuid));
     }
+    if ($response->getStatusCode() < 200 || $response->getStatusCode() >= 300) {
+      throw new \Exception(sprintf("Entity with UUID = %s cannot be deleted. Message: %s", $uuid, (string) $response->getBody()));
+    }
```
saxumVermes: We shouldn't throw an exception here in every case. For example if the client cdf is already missing, we should not halt the process just because there isn't one. This prevents the deletion of clients which don't have a corresponding client CDFs due to a purge operation. These are typically those, that are not connected to content hub anymore.

saxumVermes (changes requested):
I left a comment, we need some refactor when we delete the client.

src/ContentHubClient.php
narendradesai: Alright, I've removed exceptions thrown, unsuccessful responses will be handled in the calling methods.

src/ContentHubClient.php
```diff
@@ -1028,10 +1028,7 @@ public function unSuppressWebhook(string $webhook_uuid) {
   public function deleteClient($client_uuid = NULL) {
     $settings = $this->getSettings();
     $uuid = $client_uuid ?? $settings->getUuid();
-    $response = $this->deleteEntity($uuid);
-    if (!$response || $response->getStatusCode() < 200 || $response->getStatusCode() >= 300) {
-      throw new \Exception(sprintf("Entity with UUID = %s cannot be deleted.", $uuid));
-    }
+    $this->deleteEntity($uuid);
```
saxumVermes: In this case we wouldn't know why we were not able to delete the client CDF if there's no indication. The exception was needed I guess to check what the reason was, because we cannot return multiple responses. The only case basically when it is fine if there wasn't any client, when the service responds with a 404. Thoughts?

saxumVermes (approved):
Looks good! Thanks @narendradesai!

amangrover90 (approved):
LGTM!!

PR #184, opened 2023-11-10
LCH-6752: Add response observer method and status codes.

saxumVermes:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClientCommonTrait.php
+++ b/src/ContentHubClientCommonTrait.php
@@ -14,6 +14,13 @@ if (defined('\GuzzleHttp\ClientInterface::MAJOR_VERSION')) {
   trait ContentHubClientCommonTrait {
 
     /**
+     * The last call's response object.
+     *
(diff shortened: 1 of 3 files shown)


kirtigarg2584 (approved):

src/ContentHubClientCommonTrait.php
```diff
@@ -83,11 +111,25 @@ public function sendAsync(RequestInterface $request, array $options = []) {
...
+    /**
+     * Returns the response object from the last call.
+     *
+     * In case further examination needed e.g. status code or error message.
+     *
+     * @return \Psr\Http\Message\ResponseInterface
+     *   The response object.
+     */
+    public function getResponse(): ResponseInterface {
+      return $this->response;
```
narendradesai, line 132: This function can return `null` if $response is not initialised

PR #185, opened 2023-11-28
LCH-6798: Handle exceptions which do not have response objects.

saxumVermes:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**


narendradesai (changes requested):
In cases of missing header, we are getting exceptions like `MalformedResponseException` from HMAC library.. we can't  make strict exception types

narendradesai (approved):
Looks fine

PR #187, opened 2023-11-30
LCH-6790-2.x : Port maintenance mode changes

kirtigarg2584:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -50,6 +50,13 @@ class ContentHubClient implements ClientInterface {
   protected $settings;
 
   /**
+   * The logger responsible for tracking request failures.
+   *
+   * @var \Psr\Log\LoggerInterface
+   */
+  public $logger;
+
+  /**
    * The Event Dispatcher.
    *
    * @var \Symfony\Component\EventDispatcher\EventDispatcherInterface
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -50,13 +50,6 @@ class ContentHubClient implements ClientInterface {
   protected $settings;
 
   /**
-   * The logger responsible for tracking request failures.
-   *
-   * @var \Psr\Log\LoggerInterface
-   */
-  protected $logger;
-
-  /**
    * The Event Dispatcher.
    *
    * @var \Symfony\Component\EventDispatcher\EventDispatcherInterface
@@ -1085,7 +1078,7 @@ class ContentHubClient implements ClientInterface {
    * @return array
    *   Processed arguments.
    */
-  protected function addSearchCriteriaHeader(array $args) {
+  public function addSearchCriteriaHeader(array $args) {
     $result = explode('?', $args[0] ?? '');
     if (count($result) < 2) {
       return $args;
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -50,13 +50,6 @@ class ContentHubClient implements ClientInterface {
   protected $settings;
 
   /**
-   * The logger responsible for tracking request failures.
-   *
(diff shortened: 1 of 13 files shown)


src/ContentHubLoggingClient.php
saxumVermes, line 1: We don't need the logging client in 2.x version of the module, because it doesn't utilize the event service. That only become available in 3.x version of the module.

src/ContentHubLoggingClientInterface.php
saxumVermes, line 1: And we are not going to need this either.

src/ContentHubClientCommonTrait.php
```diff
@@ -84,6 +105,13 @@ public function getConfig(?string $option = NULL) {
    */
   trait ContentHubClientCommonTrait {
 
+    /**
+     * The last call's response object.
+     *
+     * @var \Psr\Http\Message\ResponseInterface
+     */
+    private ResponseInterface $response;
```
saxumVermes, line 113: In this version of the library we still support 7.3. That would break backward compatibility.

src/ContentHubDescriptor.php
saxumVermes, line 1: We don't need that class either, because it is only used in ContentHubLoggingClient.

src/Syndication/SyndicationEvents.php
saxumVermes, line 1: We dont' need this either, because it is related to event service.

src/Syndication/SyndicationStatus.php
saxumVermes, line 1: We can remove this too.

src/ContentHubClientCommonTrait.php
```diff
@@ -13,6 +13,13 @@
...
```
saxumVermes, line 21: In this version of the library we still support 7.3. That would break backward compatibility.

src/ContentHubClient.php
```diff
@@ -1273,4 +1266,84 @@ public function isFeatured(): bool {
...
+   *
+   * @return \Psr\Http\Message\ResponseInterface
+   *   Response object.
+   */
+  public function updateInterestListBySiteRole(string $webhook_uuid, string $site_role, array $interest_list): ResponseInterface {
+    $options['body'] = json_encode($interest_list);
+
+    return $this->put("interest/webhook/$webhook_uuid/$site_role", $options);
+  }
+
```
saxumVermes, line 1348: These can be removed, we are not using these functions. They're supposed to only exist in 3.x version.

narendradesai (changes requested):
There are unnecessary file changes in this PR. Need to refactor it. Will work on this in morning.

src/Logging/LoggingHelperTrait.php
```diff
@@ -0,0 +1,159 @@
...
+ * Helper trait for logging operations.
+ */
+trait LoggingHelperTrait {
+
+  /**
+   * The logger service.
+   *
+   * @var \Psr\Log\LoggerInterface
+   */
+  public LoggerInterface $logger;
```
saxumVermes, line 20: In this version of the library we still support 7.3. That would break backward compatibility.

saxumVermes (commented):
@kirtigarg2584 By looking through the code, tbh I think the only necessary change in this PR is to add the new request observer function through which we can retrieve the last response.

narendradesai:
@saxumVermes @kirtigarg2584 ,  I've created a separate PR for Porting maintenance mode changes to 2.x: https://github.com/acquia/content-hub-php/pull/191

PR #193, opened 2023-12-05
LCH-6798: Handle exceptions which do not have response objects. 3.4.x

narendradesai:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**


src/Logging/LoggingHelperTrait.php
```diff
@@ -35,7 +35,7 @@ trait LoggingHelperTrait {
    *  @codeCoverageIgnore
    */
   protected function getExceptionResponse(string $method, string $api_call, \Exception $exception): ResponseInterface {
```
kirtigarg2584, line 37: \Exception can be changed to RequestException and doc should be updated.

saxumVermes (approved):

PR #206, opened 2025-07-31
CHS-8047 | Remove calls to deprecated endpoints

kirtigarg2584:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -256,7 +256,7 @@ class ContentHubClient implements ClientInterface {
    * @return bool
    *   Whether the clientName from the request matches the name passed to it.
    *
-   * @deprecated in 3.7.0 and is removed from 3.8.0.
+   * @deprecated in 3.7.0 and is removed from 3.8.0. Use getClientByName().
    * @see getClientByName()
    */
   public static function clientNameExists(
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -281,7 +281,7 @@ class ContentHubClient implements ClientInterface {
     $options['body'] = json_encode(['name' => $name]);
     // Attempt to fetch the client name, if it works.
     try {
-      $client->get("settings/clients/$name");
+      $client->get("settings/client/name/$name");
 
       return TRUE;
     }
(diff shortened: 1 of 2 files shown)


saxumVermes (approved):
Let's roll it! 🔥

src/ContentHubClient.php
```diff
@@ -281,7 +281,7 @@ public static function clientNameExists(
     $options['body'] = json_encode(['name' => $name]);
     // Attempt to fetch the client name, if it works.
     try {
-      $client->get("settings/clients/$name");
+      $client->get("settings/client/name/$name");
```
narendradesai, line 287:
```suggestion
      $this->getClientByName($name);
```
We have an existing public function for this endpoint, we can reuse the same.

narendradesai (changes requested):
@kirtigarg2584 I've left one suggestion.

src/ContentHubClient.php
saxumVermes: The function is a static function, should not refer with `$this` in class internal context.

saxumVermes: Be careful of the context.

narendradesai: Ah, I missed `static` scope. Will approve it.
**Side note:** I don't see usage of this static function anywhere.

narendradesai (approved):

src/ContentHubClient.php
saxumVermes: Yeah, that's a different issue, it can be removed in that case, if it is really not used.

kirtigarg2584: As discussed, function marked as deprecated.

src/ContentHubClient.php
```diff
@@ -255,6 +255,9 @@ public static function register(
    *
    * @return bool
    *   Whether the clientName from the request matches the name passed to it.
+   *
+   * @deprecated in 3.7.0 and is removed from 3.8.0. Use getClientByName().
+   * @see getClientByName()
```
narendradesai, line 260: I think this message should be `@deprecated in 3.6.1 and ....`
If not then, PR should be made for 3.7.x  and not for 3.6.x.

@kirtigarg2584 @saxumVermes what you guys think? 🤔

PR #210, opened 2025-09-15
CHS-6426 | Implementation for Entity_deletion propagation flag updated

kirtigarg2584:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -459,7 +459,7 @@ class ContentHubClient implements ClientInterface {
    * @throws \GuzzleHttp\Exception\RequestException
    */
   public function deleteEntity(string $uuid, bool $propagate = TRUE) {
-    return $this->delete("entities/$uuid", ['propagate' => $propagate]);
+    return $this->delete("entities/$uuid", ['query' => ['propagate' => $propagate ? 'true' : 'false']]);
   }
 
   /**
@@ -477,7 +477,7 @@ class ContentHubClient implements ClientInterface {
    */
   public function deleteEntities(array $uuids, bool $propagate = TRUE) {
     $options['body'] = json_encode($uuids);
-    $options['propagate'] = $propagate;
+    $options['query'] = ['propagate' => $propagate ? 'true' : 'false'];
     return self::getResponseJson($this->delete("entities", $options));
   }
 
--- a/test/ContentHubClientTest.php
+++ b/test/ContentHubClientTest.php
@@ -876,7 +876,7 @@ class ContentHubClientTest extends TestCase {
     $this->ch_client
       ->shouldReceive('delete')
       ->once()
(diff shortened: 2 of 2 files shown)


src/ContentHubClient.php
```diff
@@ -459,7 +459,7 @@ public function postEntities(CDFObject ...$objects) {
    * @throws \GuzzleHttp\Exception\RequestException
    */
   public function deleteEntity(string $uuid, bool $propagate = TRUE) {
-    return $this->delete("entities/$uuid", ['propagate' => $propagate]);
+    return $this->delete("entities/$uuid", ['query' => ['propagate' => $propagate ? 'true' : 'false']]);
```
narendradesai, line 462:
```suggestion
    return $this->delete("entities/$uuid", ['query' => ['propagate' => $propagate]]);
```

src/ContentHubClient.php
```diff
@@ -477,7 +477,7 @@ public function deleteEntity(string $uuid, bool $propagate = TRUE) {
    */
   public function deleteEntities(array $uuids, bool $propagate = TRUE) {
     $options['body'] = json_encode($uuids);
-    $options['propagate'] = $propagate;
+    $options['query'] = ['propagate' => $propagate ? 'true' : 'false'];
```
narendradesai, line 480: is boolean value case sensitive?

narendradesai (changes requested):

src/ContentHubClient.php
kirtigarg2584: Endpoint expects string value for this parameter

src/ContentHubClient.php
kirtigarg2584: Endpoint expects string value for this parameter not boolean

saxumVermes: Where do this information coming from @kirtigarg2584? I believe that's wrong.

kirtigarg2584: I took reference from `ach-api` as mentioned in the ticket. Also, with boolean values function is not working as expected.

Screenshot 2025-09-18 at 6 56 55 PM

saxumVermes: I'm confused, why is it not working as expected?

saxumVermes: as it turns out `http_build_query` converts boolean values into integers which of course results in a wrong query string. For reference the function is used to convert the request options, query, into query string under the hood.

saxumVermes (approved):

src/ContentHubClient.php
narendradesai: I see, thanks @saxumVermes

narendradesai (approved):

PR #213, opened 2025-09-29
CHS-8839: Added fetch entities function.

narendradesai:
**Motivation**

Fixes https://acquia.atlassian.net/browse/CHS-8839

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1589,12 +1589,12 @@ class ContentHubClient implements ClientInterface {
   }
 
   /**
-   * Fetches entities from service queue for processing.
+   * Receives queue items from service queue for processing.
    *
    * @param int $limit
-   *   Maximum number of items to return.
+   *   Maximum number of queue items to return.
    * @param string $visibility_timeout
-   *   Duration for which the fetched entities will be invisible to other
+   *   Duration for which the fetched queue items will be invisible to other
    *   queue consumers. Must be suffixed with duration unit (m, s, h, d, etc).
    * @param array $queue_filters
    *   An array of queue filters to apply. Supported values are:
@@ -1604,11 +1604,11 @@ class ContentHubClient implements ClientInterface {
    *   If not specified, queued items will be returned.
    *
    * @return array|null
-   *   Response from backend call.
+   *   Response from backend call, that contains array of queue items.
    *
    * @throws \Exception
    */
-  public function fetchEntities(int $limit, string $visibility_timeout, array $queue_filters = []): ?array {
+  public function receiveQueueItems(int $limit, string $visibility_timeout, array $queue_filters = []): ?array {
     $config = $this->getConfig();
     $config['headers']['X-Acquia-Content-Hub-Syndication'] = 'ReceiveQueueItems';
     $this->setConfigs($config);
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1614,16 +1614,15 @@ class ContentHubClient implements ClientInterface {
     $this->setConfigs($config);
 
     $options = [
-      'body' => json_encode([
+      'body' => [
         'max_number_of_items' => $limit,
         'visibility_timeout' => $visibility_timeout,
-      ]),
+      ],
     ];
     if (!empty($queue_filters)) {
-      $body = json_decode($options['body'], TRUE);
-      $body['queue_filters']['state'] = $queue_filters;
-      $options['body'] = json_encode($body);
+      $options['body']['queue_filters']['state'] = $queue_filters;
     }
+    $options['body'] = json_encode($options['body']);
     return self::getResponseJson($this->post('queues/syndications', $options));
   }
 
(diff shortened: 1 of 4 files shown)


test/ContentHubClientTest.php
```diff
@@ -3301,4 +3301,83 @@ public function testGetResponse(): void {
...
 
+  /**
+   * Tests fetchEntities.
+   *
+   * @covers::fetchEntities
+   */
+  public function testFetchEntities(): void {
+    $limit = 10;
+    $visibility_timeout = '1d';
+    $queue_filters = ['queued'];
```
kirtigarg2584, line 3444: limit, visibility_timeout and queue_filters can be directly set in expected_request_body. These variables can be removed.

narendradesai: Those variables are used at multiple places. Any reason why we should consider removing them?

kirtigarg2584: They are used at total 2 places. For `expected_request_body` directly values can be assigned and at 3361 values from `expected_request_body` can be used.

narendradesai: It is alright to use variables, when used more than once.

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,43 @@ public function isFeatured(): bool {
...
+
+    $options = [
+      'body' => json_encode([
+        'max_number_of_items' => $limit,
+        'visibility_timeout' => $visibility_timeout,
+      ]),
+    ];
+    if (!empty($queue_filters)) {
+      $body = json_decode($options['body'], TRUE);
+      $body['queue_filters']['state'] = $queue_filters;
```
kirtigarg2584: This processing can be optimised.

$body = [
  'max_number_of_items' => $limit,
  'visibility_timeout' => $visibility_timeout,
];
if (!empty($queue_filters)) {
  $body['queue_filters']['state'] = $queue_filters;
}
$options['body'] = json_encode($body);

In the current code, $options is encoded and then decoded in-case the if condition is satisfied and then encoded again. Instead let's finalize the $options and then encode it once.

narendradesai: Thanks @kirtigarg2584 , updated the code.

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,42 @@ public function isFeatured(): bool {
...
+   *   - 'queued'
+   *   - 'processing'
+   *   If not specified, queued items will be returned.
+   *
+   * @return array|null
+   *   Response from backend call.
+   *
+   * @throws \Exception
+   */
+  public function fetchEntities(int $limit, string $visibility_timeout, array $queue_filters = []): ?array {
```
saxumVermes: Please use `fetchQueueItems`

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,42 @@ public function isFeatured(): bool {
...
+   *   queue consumers. Must be suffixed with duration unit (m, s, h, d, etc).
+   * @param array $queue_filters
+   *   An array of queue filters to apply. Supported values are:
+   *   - 'failed'
+   *   - 'queued'
+   *   - 'processing'
+   *   If not specified, queued items will be returned.
+   *
+   * @return array|null
+   *   Response from backend call.
```
saxumVermes: Please use a more descriptive text

src/ContentHubClient.php
saxumVermes: on the other hand you can just call this function `receiveQueueItems`, since that's the action you're sending to the service which is even better.

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,42 @@ public function isFeatured(): bool {
...
   }
 
+  /**
+   * Receives queue items from service queue for processing.
+   *
+   * @param int $limit
+   *   Maximum number of queue items to return.
+   * @param string $visibility_timeout
+   *   Duration for which the fetched queue items will be invisible to other
+   *   queue consumers. Must be suffixed with duration unit (m, s, h, d, etc).
```
saxumVermes: day is not a valid one though: `"ns", "us" (or "µs"), "ms", "s", "m", "h"`.

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,42 @@ public function isFeatured(): bool {
...
+   * @param array $queue_filters
+   *   An array of queue filters to apply. Supported values are:
+   *   - 'failed'
+   *   - 'queued'
```
saxumVermes, line 1604: only these two are returned through this api

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,42 @@ public function isFeatured(): bool {
...
+   *
+   * @return array|null
+   *   Response from backend call, that contains array of queue items.
+   *
+   * @throws \Exception
+   */
+  public function receiveQueueItems(int $limit, string $visibility_timeout, array $queue_filters = []): ?array {
+    $config = $this->getConfig();
+    $config['headers']['X-Acquia-Content-Hub-Syndication'] = 'ReceiveQueueItems';
+    $this->setConfigs($config);
```
saxumVermes: no need for this, it can be set through:

```php
    $options['headers'] = // ... rest 
```

saxumVermes (approved):

kirtigarg2584 (approved):
Looks good 🎉

PR #214, opened 2025-10-03
CHS-8848 | Drush ach-qs:delete Command

kirtigarg2584:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1513,31 +1513,83 @@ class ContentHubClient implements ClientInterface {
   }
 
   /**
-   * Get entities from service queue.
+   * Provides an array of queue items.
    *
    * @return array
-   *   Response from backend call.
+   *   Returns array of queue items from cloud queue with count.
+   *   Format:
+   *   [
+   *    'total' => 3,
+   *    'success' => true,
+   *    'data' => [
+   *     [
+   *       'id' => '1',
+   *       'entity_uuid' => '5f71af85-cbb0-48ac-84f3-97083bf16367',
+   *       'client_uuid' => '8fb4c61c-bc0c-4451-a1aa-f576bf4eb966',
+   *       'state' => 'queued',
+   *       'payload' => [
+   *         'action' => 'entity_create'
+   *       ],
+   *     'visible_at' => '1753879023',
+   *     'created_at' => '1753879023',
+   *     'updated_at' => '1753879023'
+   *     ],
+   *    ]
+   *  ]
    *
    * @throws \Exception
    */
-  public function getAllEntitiesFromServiceQueue(): array {
+  public function getAllQueueItems(): array {
     return self::getResponseJson($this->get('queues/syndications'));
   }
 
   /**
-   * Deletes entities from service queue.
-   *
-   * @param array $options
-   *   If empty, all items will be deleted.
-   *   Otherwise comma-separated specific ids should be sent
-   *   as $options['syndication_ids'] = '1,2' for deletion.
+   * Deletes all items from service queue.
    *
    * @return array|null
-   *   Response from backend call.
+   *   Response from Pull Syndication API.
    *
    * @throws \Exception
    */
-  public function deleteEntitiesFromServiceQueue(array $options = []): ?array {
+  public function purgeQueue(): ?array {
+    return self::getResponseJson($this->delete('queues/syndications'));
+  }
+
+  /**
+   * Deletes items from service queue using the specified syndication_ids.
+   *
+   * @param array $options
+   * <code>
+   *  $options = [
+   *      "syndication_ids": [1,2,3,4,5],
+   *  ];
+   *  </code>
+   *
+   * @return array|null
+   *   Response from Pull Syndication API.
+   *
+   * @throws \Exception
+   */
+  public function deleteQueueItemsBySyndicationIds(array $options = []): ?array {
+    return self::getResponseJson($this->delete('queues/syndications', $options));
+  }
+
+  /**
+   * Deletes items from service queue using the specified entity_uuids.
+   *
+   * @param array $options
+   * <code>
+   *  $options = [
+   *      "entity_uuids": ["1c6f7a8a-5cf6-4bb2-8e09-e64c0d63629d","4c6f7a8a-5cf6-4bb2-8e09-e64c0d63629e"]
+   *  ];
+   *  </code>
+   *
+   * @return array|null
+   *   Response from Pull Syndication API.
+   *
+   * @throws \Exception
+   */
+  public function deleteQueueItemsByEntityUuids(array $options = []): ?array {
     return self::getResponseJson($this->delete('queues/syndications', $options));
   }
 
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1548,7 +1548,7 @@ class ContentHubClient implements ClientInterface {
    * Deletes all items from service queue.
    *
    * @return array|null
-   *   Response from Pull Syndication API.
+   *   Response from Syndication Queue API.
    *
    * @throws \Exception
    */
@@ -1563,7 +1563,7 @@ class ContentHubClient implements ClientInterface {
    *   Array with syndication IDs of items to be removed.
    *
    * @return array|null
-   *   Response from Pull Syndication API.
+   *   Response from Syndication Queue API.
    *
    * @throws \Exception
    */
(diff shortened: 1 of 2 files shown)


src/ContentHubClient.php
```diff
@@ -1512,4 +1512,42 @@ public function isFeatured(): bool {
...
+  public function deleteEntitiesFromServiceQueue(array $options): ?array {
+
+    if (isset($options['all'])) {
+      return self::getResponseJson($this->delete('queues/syndications'));
+    }
+
+    if (isset($options['ids'])) {
+      $body['ids'] = $options['ids'];
+      return self::getResponseJson($this->delete('queues/syndications', $body));
+    }
```
narendradesai: I think logic of transforming options, should be handled inside module.
IMO, library functions should be kept as clean and simple as possible.

Also, endpoint does not expect `ids` key, instead it expects `syndication_ids` key. Please refer: https://acquia.atlassian.net/browse/CHS-8862

narendradesai (changes requested):

src/ContentHubClient.php
kirtigarg2584: function updated

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,33 @@ public function isFeatured(): bool {
...
 
+  /**
+   * Get entities from service queue.
+   *
+   * @return array
+   *   Response from backend call.
+   *
+   * @throws \Exception
+   */
+  public function getAllEntitiesFromServiceQueue(): array {
```
saxumVermes:
```suggestion
  public function getAllQueueItems(): array {
```

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,33 @@ public function isFeatured(): bool {
...
+   *   If empty, all items will be deleted.
+   *   Otherwise comma-separated specific ids should be sent
+   *   as $options['syndication_ids'] = '1,2' for deletion.
+   *
+   * @return array|null
+   *   Response from backend call.
+   *
+   * @throws \Exception
+   */
+  public function deleteEntitiesFromServiceQueue(array $options = []): ?array {
```
saxumVermes: Please create a new function, `emptyQueue` or `purgeQueue` which will delete all items from the queue. We can also call `deleteQueue`. I think `purgeQueue` would make the most sense.
Let's not mash up multiple behaviours in the same method.

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,33 @@ public function isFeatured(): bool {
...
+    return self::getResponseJson($this->get('queues/syndications'));
+  }
+
+  /**
+   * Deletes entities from service queue.
+   *
+   * @param array $options
+   *   If empty, all items will be deleted.
+   *   Otherwise comma-separated specific ids should be sent
+   *   as $options['syndication_ids'] = '1,2' for deletion.
```
saxumVermes: Here we can break this down into multiple methods:
* `deleteQueueItemsBySyndicationIds`
* `deleteQueueItemsByEntityUuids`
* and the third one I've already mentioned at https://github.com/acquia/content-hub-php/pull/214/files#r2411354859: `purgeQueue`

saxumVermes (changes requested):

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,33 @@ public function isFeatured(): bool {
     return $remote['featured'] ?? FALSE;
   }
 
+  /**
+   * Get entities from service queue.
+   *
+   * @return array
+   *   Response from backend call.
```
saxumVermes: This comment doesn't tell too much, please use something concrete such as `A list of queue items` perhaps you can also be a bit more  concrete with the structure of the array or even include it in the comments.

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,33 @@ public function isFeatured(): bool {
...
+  /**
+   * Deletes entities from service queue.
+   *
+   * @param array $options
+   *   If empty, all items will be deleted.
+   *   Otherwise comma-separated specific ids should be sent
+   *   as $options['syndication_ids'] = '1,2' for deletion.
+   *
+   * @return array|null
+   *   Response from backend call.
```
saxumVermes: Same here.

src/ContentHubClient.php
kirtigarg2584: What's the use of creating multiple functions here? All of them will use the same endpoint in the same manner. Processing of the $options is already done in the corresponding Drush method.

saxumVermes: The following can be written up as the reason for creating multiple specific functions:
* Separation of concern
* Cleaner library and api design
* More communicative behaviour, signifies intent
* Increased maintainability
* Enforcing a better terminology and language which is used in the business layer as well: ubiquities language
* Better testability as they are smaller in size and only target one behaviour
* etc.

That's just a few off the top of my head, but surely there can be many more. These are pretty generic so can fit to other use cases as well.

src/ContentHubClient.php
kirtigarg2584: Done

src/ContentHubClient.php
kirtigarg2584: Done

src/ContentHubClient.php
kirtigarg2584: Done

src/ContentHubClient.php
kirtigarg2584: Done

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,92 @@ public function isFeatured(): bool {
...
+   *   'success' => true,
+   *   'data' => [
+   *     [
+   *       'id' => '1',
+   *       'entity_uuid' => '5f71af85-cbb0-48ac-84f3-97083bf16367',
+   *       'client_uuid' => '8fb4c61c-bc0c-4451-a1aa-f576bf4eb966',
+   *       'state' => 'queued',
+   *       'payload' => [
+   *       'action' => 'entity_create'
+   *     ],
```
saxumVermes, line 1534: this should be removed

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,92 @@ public function isFeatured(): bool {
...
+   *
+   * @param array $options
+   *   Query options.
+   *
+   * @return array|null
+   *   Response from Pull Syndication API.
+   *
+   * @throws \Exception
+   */
+  public function deleteQueueItemsBySyndicationIds(array $options = []): ?array {
```
saxumVermes: In this case instead of options, it should accept an array of syndication ids, integers

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,92 @@ public function isFeatured(): bool {
...
+    return self::getResponseJson($this->delete('queues/syndications', $options));
```
saxumVermes, line 1572: The option should be initialized:
```php
$options['syndication_ids'] = $syndication_ids // int[]
```

src/ContentHubClient.php
```diff
@@ -1512,4 +1512,92 @@ public function isFeatured(): bool {
...
+   *
+   * @param array $options
+   *   Query options.
+   *
+   * @return array|null
+   *   Response from Pull Syndication API.
+   *
+   * @throws \Exception
+   */
+  public function deleteQueueItemsByEntityUuids(array $options = []): ?array {
```
saxumVermes: similar to https://github.com/acquia/content-hub-php/pull/214/files#r2414001774

saxumVermes: otherwise there would be no difference between the two functions

src/ContentHubClient.php
kirtigarg2584: Removed

saxumVermes: rather correctly indented

kirtigarg2584: Indentation fixed

src/ContentHubClient.php
kirtigarg2584: Done

src/ContentHubClient.php
kirtigarg2584: Done

src/ContentHubClient.php
kirtigarg2584: Done

saxumVermes (approved):

PR #222, opened 2025-11-05
CHS-8899: Implement confirm processed queue api.

saxumVermes:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

### DIFF
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1725,7 +1725,7 @@ class ContentHubClient implements ClientInterface {
    *   An array of entity UUIDs that failed processing.
    *
    * @return array|null
-   *   The response array
+   *   The response array.
    *
    * @throws \Exception
    */
--- a/src/ContentHubClient.php
+++ b/src/ContentHubClient.php
@@ -1716,6 +1716,19 @@ class ContentHubClient implements ClientInterface {
     return self::getResponseJson($this->post('queues/syndications', $options));
   }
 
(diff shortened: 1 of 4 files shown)


narendradesai (approved):

src/ContentHubClient.php
```diff
@@ -1716,4 +1716,34 @@ public function receiveQueueItems(int $limit, string $visibility_timeout, array
...
+   * @throws \Exception
+   */
+  public function confirmProcessedQueueItems(array $processed = [], array $failed = []): ?array {
+    $options = [
+      'headers' => [
+        SyndicationQueue::HEADER => SyndicationQueue::CONFIRM_PROCESSED_QUEUE_ITEMS,
+      ],
+    ];
+    if (!empty($processed)) {
+      $options['body']['entity_uuids']['processed'] = $processed;
```
kirtigarg2584: I think we should use `RequestOptions::BODY` here.

src/ContentHubClient.php
```diff
@@ -1716,4 +1716,34 @@ public function receiveQueueItems(int $limit, string $visibility_timeout, array
...
+    }
+    if (!empty($failed)) {
+      $options['body']['entity_uuids']['failed'] = $failed;
```
kirtigarg2584: same as above, use `RequestOptions::BODY` here.

